PR #1, opened 2018-10-30
Add shallow submodules

nickdesaulniers:
Added with:
$ git submodule add --depth 1 -- \
    git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git
$ git submodule add --depth 1 -- https://git.llvm.org/git/llvm.git/
$ git submodule add --depth 1 -- https://git.llvm.org/git/clang.git/
$ git submodule add --depth 1 -- git://git.buildroot.net/buildroot

Forced shallow with:
$ git config -f .gitmodules submodule.linux.shallow true
$ git config -f .gitmodules submodule.llvm.shallow true
$ git config -f .gitmodules submodule.clang.shallow true
$ git config -f .gitmodules submodule.buildroot.shallow true

Will need to figure out a symlink for clang though, if we can even build
it out of tree? :man_shrugging:


.gitmodules
```diff
@@ -0,0 +1,16 @@
+[submodule "linux"]
+	path = linux
+	url = git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git
+	shallow = true
+[submodule "llvm"]
+	path = llvm
+	url = https://git.llvm.org/git/llvm.git/
+	shallow = true
+[submodule "clang"]
+	path = clang
```
nathanchance: llvm/tools/clang wouldn't work?

driver.sh
```diff
@@ -34,12 +33,28 @@ build_linux() {
...
 }
 
+build_root() {
+  mkdir -p buildroot/overlays/etc/init.d/
+  cp -f inittab buildroot/overlays/etc/.
+  cp -f S50yolo buildroot/overlays/etc/init.d/.
+
+  cd buildroot
+  make defconfig BR2_DEFCONFIG=../buildroot.config
+  make
```
nathanchance: probably want to `make -j$(nproc)` (GitHub hates the ticks in markdown lol) this too just for consistency.

A make wrapper might not be a bad idea either

```
function make() {
  command make -j`nproc` "$@"
}
```

S50yolo
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cat /proc/version
+reboot
```
nathanchance: If we just want to shutdown the machine, this could be `poweroff` and just remove the `-no-reboot` QEMU flag but I haven't actually tested yet.

driver.sh
```diff
@@ -0,0 +1,82 @@
...
+  test -x $clang
+  set +e
+
+  cd linux
+  export ARCH=arm64
+  export CROSS_COMPILE=aarch64-linux-gnu-
+  # TODO: do we really want to be building the kernel from scratch every time?
+  # Rerunning ninja above will not rebuild unless source files change.
+  #make CC=$clang mrproper
+  make CC=$clang defconfig
```
nathanchance: HOSTCC too?

driver.sh
```diff
@@ -0,0 +1,82 @@
...
+
+  set +e
+}
+
+build_clang() {
+  ln -sf ../../clang llvm/tools/clang
+  mkdir -p llvm/build
+  cd llvm/build
+
+  cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ..
```
nathanchance: Might want to look at some of the threading options like `LLVM_PARALLEL_COMPILE_JOBS` and `LLVM_PARALLEL_LINK_JOBS` since Travis timed out (I didn't see an error so I just assume it wasn't going fast enough but don't have time to read [this](https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out)). Can also probably ccache it up (`LLVM_CCACHE_BUILD`)

.gitmodules
nickdesaulniers: No.  Can't add a submodule in the path of another submodule in git.

driver.sh
nickdesaulniers: There's no rendering issue with github's markdown renderer.  I guess I'm curious if folks using bash have an issue with the backticks?  No, seems like backticks work there.

Does buildroot support `-j`?

.gitmodules
stephenhines: You can use LLVM_EXTERNAL_{CLANG,LLD,POLLY}_SOURCE_DIR (look on https://llvm.org/docs/CMake.html for more info). These variables allow you to point to various projects that aren't in the usual nested location.

S50yolo
nickdesaulniers: That's a good idea. I was thinking of adding that last night but was simply too tired.  Will try it out.

driver.sh
nickdesaulniers: Good idea.  Will add.

driver.sh
nickdesaulniers: Talking with some teammates, their ideas were to disable other backends that we don't need.  I consider that an antipattern for the distribution of clang, but to speed up our tests, it may be fine.

Travis also had some documentation I recall on using ccache, then caching that cache dir between builds.

Oh, looks like there are nightly builds of clang? https://apt.llvm.org/  That would save a lot of time!

.gitmodules
```diff
@@ -0,0 +1,16 @@
...
+	url = https://git.llvm.org/git/clang.git/
+	shallow = true
+[submodule "buildroot"]
+	path = buildroot
+	url = git://git.buildroot.net/buildroot
+	shallow = true
```
nickdesaulniers: interesting, if I try to clone this locally, I get an error that the buildroot server doesn't advertise the commit listed.  travis ci doesn't see this because it always clones with --depth 50.  I wonder if I should remove the shallow clone here, or update the readme on how to initialize the repo?

driver.sh
nathanchance: Ah ```make -j`nproc` ``` works

Shellcheck recommends using `$(...)` over backticks and that's what I always use personally: https://github.com/koalaman/shellcheck/wiki/SC2006

I thought it did but maybe not.

driver.sh
nathanchance: I think switching to the nightly builds is ideal so that we don't have to tax Travis lol.

.gitmodules
nickdesaulniers: removed buildroot as a submodule.  I *think* this won't be an issue with the kernel if we wire up some kind of auto update.

driver.sh
```diff
@@ -0,0 +1,78 @@
...
+  fi
+
+  if [[ ! -e buildroot/output/images/rootfs.ext4 ]]; then
+    mkdir -p buildroot/overlays/etc/init.d/
+    cp -f inittab buildroot/overlays/etc/.
+    cp -f S50yolo buildroot/overlays/etc/init.d/.
+
+    cd buildroot
+    make defconfig BR2_DEFCONFIG=../buildroot.config
+    mako_reactor
```
nickdesaulniers: :man: even buildroot takes too long.  Might have to make these prebuilt images, too...

driver.sh
nickdesaulniers: done

driver.sh
nickdesaulniers: done

.gitmodules
nathanchance: I've added 4eba2e8bbcc86cfdf4db7abf7f7cb948be1567cb + added a [Travis cron](https://docs.travis-ci.com/user/cron-jobs/) for now but I could wire up a cron job on my server to automatically bump and commit the update every couple of hours like I am doing for the other CBL repos.

driver.sh
nathanchance: I've uploaded prebuilts in https://github.com/ClangBuiltLinux/continuous-integration/commit/aeea4e2b3c3e85adcee912ebdd8182fc6dffeb5f

S50yolo
nathanchance: It does work, added in https://github.com/ClangBuiltLinux/continuous-integration/commit/aeea4e2b3c3e85adcee912ebdd8182fc6dffeb5f

nathanchance:
Actually looks like that updated QEMU fixed an issue with arm kernels failing to shutdown properly: https://travis-ci.com/ClangBuiltLinux/continuous-integration/builds/89775252

Time for bed now that both [arm](https://travis-ci.com/ClangBuiltLinux/continuous-integration/builds/89776717) and [arm64](https://travis-ci.com/ClangBuiltLinux/continuous-integration/builds/89784328) are passing, feel free to check out my nc branch and see what you think. I'll clean it up a bit when I wake up.

Side note, I think we might need one branch for each architecture since I don't think there is any way we're getting more than one kernel build out of each run.

Lastly, troll of the night:

```
remote: warning: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
remote: warning: See http://git.io/iEPt8g for more information.
remote: warning: File images/arm/rootfs.ext4 is 60.00 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
remote: warning: File images/arm64/rootfs.ext4 is 60.00 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
To github.com:ClangBuiltLinux/continuous-integration.git
   981ab90..e5d838d  nc -> nc
nathan@flashbox ~/cbl/ci (nc) $ du -sh images/arm/rootfs.ext4
6.1M    images/arm/rootfs.ext4
nathan@flashbox ~/cbl/ci (nc) $ du -sh images/arm64/rootfs.ext4
12M     images/arm64/rootfs.ext4
```

nickdesaulniers:
Let's get something landed for arm64, then follow up in subsequent PRs on how to enable additional arches.  Some testing per PR is vastly superior to no testing.

nickdesaulniers:
Thanks for polishing off this PR.  Github's UI won't let me approve my own PR, which makes sense, but @nathanchance or @stephenhines has to accept this revision.  It LGTM

PR #12, opened 2018-11-01
ci: Remove linux submodule

nathanchance:
At a certain point, Travis can no longer init submodules that aren't up
to date. Move the repo management into the script.

Fixes #8.

PR #13, opened 2018-11-01
buildroot: Cap ext4 images at 20M

nathanchance:
GitHub complains because the images are normally 60MB (even though the
actual disk space is much less). Lower it to 20M so that they still have
a little space to grow if ever necessary and GitHub will shut up.


nathanchance:
Ugh, I'll rebase once #12 is merged so Travis can run.

buildroot/arm64.config
```diff
@@ -10,4 +10,5 @@ BR2_PACKAGE_BASH=y
 BR2_TARGET_ROOTFS_CPIO=y
 BR2_TARGET_ROOTFS_EXT2=y
 BR2_TARGET_ROOTFS_EXT2_4=y
+BR2_TARGET_ROOTFS_EXT2_SIZE="20M"
```
nickdesaulniers: what does this config do?  Is the disk image just empty at the end?

nickdesaulniers:
thanks @nathanchance !

PR #18, opened 2018-11-03
Add x86_64 and add some small cleanup

nathanchance:
@nickdesaulniers if you want to merge your LLD stuff first, let me know and I'll rebase this.


driver.sh
```diff
@@ -78,7 +86,7 @@ check_dependencies() {
 }
 
 mako_reactor() {
-  make -j"$(nproc)" CC="$ccache $clang" HOSTCC="$ccache $clang" "$@"
+  make -j"$(nproc)" CC="${ccache} ${clang}" HOSTCC="${ccache} ${clang}" "$@"
```
nickdesaulniers: Just curious if the `$@` should be brace expanded as well? Ie `"${@}"`?

nathanchance: Done.

PR #19, opened 2018-11-03
add lld to the testing matrix

nickdesaulniers:
Probably will need to rebase once #18 lands.

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -18,6 +18,8 @@ setup_variables() {
         echo
         echo "   ARCH:"
         echo "       If no ARCH value is specified, arm64 is the default. Currently, arm, arm64, and x86_64 are supported."
+        echo "   LD:"
+        echo "       If no LD value is specified, \${CROSS_COMPILE}-ld is used. arm64 only."
         echo
         echo " Optional parameters:"
         echo "   -c | --clean:"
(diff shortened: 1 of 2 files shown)


nathanchance:
Yeah, will require a rebase because of the `mako_reactor` line (plus would allow a squash of everything). The diff looks good to me though so I'll approve this now and once #18 is merged and this is rebased, I'll add it again. Hopefully we can get lld working for other architectures soon!

driver.sh
```diff
@@ -81,12 +82,13 @@ check_dependencies() {
...
   command -v unbuffer
   command -v clang-8
+  command -v "${LD:="${CROSS_COMPILE:-}"ld}"
 
   set +e
 }
 
 mako_reactor() {
-  make -j"$(nproc)" CC="${ccache} ${clang}" HOSTCC="${ccache} ${clang}" "${@}"
+  make -j"$(nproc)" CC="${ccache} ${clang}" HOSTCC="${ccache} ${clang}" LD="${LD}" ${@}"
```
nathanchance: Whoops, a quotation mark got lost it seems.

nickdesaulniers: back to first grade for me! :ice_cream: :horse_racing:

PR #22, opened 2018-11-05
Use KBUILD_BUILD variables and add '-j' flag

nathanchance:


nickdesaulniers:
Thanks for implementing this feature!  Getting the build time down will allow for quicker go/no go signals on PRs! We should prioritize that! 🐎 ⚡️ 

Next time, please put separate features in separate branches and PRs.  `-j` should be in a separate pull request.  That way it's destiny is not cotangled with ccache.

It looks like ccache is now possibly not working (build 146), so that makes me pause when thinking of approving, since build 145 should have primed the cache.  I do think something is better than nothing, but with build 146 it seems that we're back to nothing, though we've added more code. (and now build 147 from the cron job has started)

Maybe we should take this, and if we don't see caching improvements with X builds, revert ccache support?

Maybe in a follow up, we can print `ccache -s` before and after compilation?  I wonder if travis is smart enough to use separate caches per matrix, and if we need to increase ccache's cache size?

nathanchance:
> Next time, please put separate features in separate branches and PRs. `-j` should be in a separate pull request. That way its destiny is not cotangled with ccache.

Sure. I figured it was benign enough to go in together but I can definitely separate things in the future.

> It looks like ccache is now possibly not working (build 146), so that makes me pause when thinking of approving, since build 145 should have primed the cache. I do think something is better than nothing, but with build 146 it seems that we're back to nothing, though we've added more code. (and now build 147 from the cron job has started)

I think the ccache caching is per branch since build 145 showed quick builds for arm32 and x86_64, which would correspond to build 143. Ultimately, this should be helpful when we stop having so much development and only rely on the cron jobs on master.

> Maybe we should take this, and if we don't see caching improvements with X builds, revert ccache support?

I think ccache support is still good to have for local development. I doubt removing it will make us see much gain.

> Maybe in a follow up, we can print `ccache -s` before and after compilation?

I can push this for this pull if you want?

```diff
diff --git a/.travis.yml b/.travis.yml
index 86f45af..5ac6041 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,8 @@ before_script:
   - ./configure
   - make -j$(nproc)
   - export PATH=${PWD}:${PATH}
+  - ccache -s
   - cd "${OLDPWD}"
 after_script:
   - sleep 1
+  - ccache -s
```

> I wonder if travis is smart enough to use separate caches per matrix, and if we need to increase ccache's cache size?

We should in theory have 18GB of space according to Travis so I think increasing the cache size to 10GB is probably worthwhile.

nickdesaulniers:
> I think ccache support is still good to have for local development. I doubt removing it will make us see much gain.

I don't think we'll be doing much local development with this repo; ccache support is more valuable in travis than locally, IMO.

> We should in theory have 18GB of space according to Travis so I think increasing the cache size to 10GB is probably worthwhile.

I think ccache defaults to some insanely low size, and I think resizing it to its existing size doesn't clear the cache, so we could add that unconditionally.

> I can push this for this pull if you want?

Followup PR, let's land this 🛩

Issue #16, opened 2018-11-02
add PPC support


nickdesaulniers:
Hey @shenki , do you have a set of configs that we can add for PPC? Probably can be modeled on @nathanchance 's patches in #14 ?

nathanchance:
2b76ea23aaca9813317c64d4239cb5f0e032aea1 too

nickdesaulniers:
cc @daxtens

daxtens:
So you want to distinguish between 3 different flavours of PowerPC: 64-bit LE, 64-bit BE and 32-bit. @shenki has done lots of work for the first and third, and I've done a little bit on the second. I'm mostly working from a particularly slimmed down config for the paticular dev board I have, so that's probably not what you want to build from. You probably want ppc64_defconfig with whatever bits are still broken disabled.

shenki:
WIP: https://github.com/shenki/continuous-integration/tree/ppc64le

Currently it contains hand-built binaries to get the build running on the outdated base distro.

We need a better solution for bringing in newer binutils and qemu. I've had success in the past using Travis but do the entire build inside a container running an up to date distro.

nathanchance:
Yeah, having access to a rolling release distro like Arch would be really nice so all of our packages are up to date. That would be nice for arm64 so we could test LSE atomics support. I'll try to take a look at this today.

nathanchance:
Hmmm actually we'll want to stick with Debian/Ubuntu given that the LLVM/Clang nightlies are only available on there. Stretch seems like a good candidate since it has an official Docker image: https://hub.docker.com/r/_/debian/

@shenki would [binutils 2.28](https://packages.debian.org/stretch/binutils-powerpc-linux-gnu) be good enough? I suppose if we needed the latest, we could just use Debian unstable (sid). Stretch also has [QEMU 2.8](https://packages.debian.org/stretch/qemu-system-ppc).

Only downside I see with switching to Docker is the ability to cache ccache (I think, unless we want to keep using the same docker image).

nickdesaulniers:
If docker simplifies things, let's do that. I just have no experience with
it, but we can fix that.

nathanchance:
I've picked it up fairly quickly. Luckily, I had a GCE trial still so I could have access to a Trusty enviroment.

What I like about it is we get a totally deterministic environment every time. This should work for Travis I think (will push to test here shortly).

```
[...]

Issue #23, opened 2018-11-05
ccache misbehavior


nickdesaulniers:
followup from #22 .

ccache has been misbehaving:
- hard to tell if other branch builds are related or not
- increase cache size
- arm64 seems to have issues with ccache, but reportedly not locally, only remote on travis.

[travis ci ccache docs](https://docs.travis-ci.com/user/caching/)

tpimh:
Why are we building ccache from source? Why not just install it from official Ubuntu repos or PPA? Is there any particular reason for this?

nathanchance:
3.1.9 (Trusty's version) has a bug where a ton of warnings were emitted, which was fixed in 3.2:

"ccache no longer passes preprocessor options like -D and -I to the compiler when compiling preprocessed output. This fixes warnings emitted by Clang.".

It takes us literally five seconds to fetch and build ccache, it's like 10 files big. I don't use Ubuntu so

tpimh:
Thanks for your explanation. Maybe you could consider using [this PPA](https://launchpad.net/~clearpath-robotics/+archive/ubuntu/ccache). But since it's just a matter of 5 seconds, it's not that important.

nickdesaulniers:
If we move to docker in #25 , is it possible to move the ccache dir to [a dir that travis itself caches](https://docs.travis-ci.com/user/caching#arbitrary-directories)? (Doesn't need to be in #25 )

nathanchance:
Yes we could. I already have to make adjustments to the PR so I might as well do it at the same time.

nathanchance:
Alright seems like it is working a little better now: https://travis-ci.com/nathanchance/continuous-integration/builds/90440872

I'll push to the PR here shortly once I clean up my testing.

nickdesaulniers:
Hey that's great! Looks like no spooky 👻 bugs with arm64 builds we were seeing before!

nickdesaulniers:
Fixes in #25 , thanks @nathanchance !  Next build (163) of master should have a cold cache, but subsequent builds should have a nice hot cache.

PR #25, opened 2018-11-05
ci: Switch to Docker

nathanchance:
Unfortunately, as we add new architectures, we are going to need more up
to date binaries that Trusty cannot provide (see #16). Move our building
infrastructure to Docker, which gives us a deterministic, configurable
environment every time. The only forseeable downside is no cacheable
ccache directory so slower builds but we are getting to a point where
constant development is slowing down. I have kept ccache support in
driver.sh so that anyone who uses it locally can still take advantage of
it.

Initially, we'll start with the latest Debian stable but if we need to
move to the unstable version (sid) for more up to date packages, it's as
easy as changing 'debian' to 'debian:sid' in .travis.yml and adjusting
the LLVM/Clang apt repo in debian-setup.sh.

### DIFF
--- a/debian-setup.sh
+++ b/debian-setup.sh
@@ -14,22 +14,23 @@ apt-get upgrade -y
 # is separate from the Clang/lld installation below
 # because we would need to at least install curl to
 # get LLVM's apt key
-apt-get install -y bc \
-                   binutils \
-                   binutils-aarch64-linux-gnu \
-                   binutils-arm-linux-gnueabi \
-                   bison \
-                   ccache \
-                   curl \
-                   flex \
-                   expect \
-                   git \
-                   gnupg \
-                   libssl-dev \
-                   openssl \
-                   make \
-                   qemu-system-arm \
-                   qemu-system-x86
+apt-get install -y \
+    bc \
+    binutils \
+    binutils-aarch64-linux-gnu \
+    binutils-arm-linux-gnueabi \
+    bison \
+    ccache \
+    curl \
+    expect \
+    flex \
+    git \
+    gnupg \
+    libssl-dev \
+    make \
+    openssl \
+    qemu-system-arm \
+    qemu-system-x86
 
 # Install nightly verisons of Clang and lld (apt.llvm.org)
 curl https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
--- a/debian-setup.sh
+++ b/debian-setup.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 # Installs all of the necessary packages in a Docker container
-# MUST BE RUN WITH ROOT
 
 # Show all commands and exit upon failure
 set -eux
--- a/debian-setup.sh
+++ b/debian-setup.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+# Installs all of the necessary packages in a Docker container
+# MUST BE RUN WITH ROOT
+
+apt-get update
+apt-get upgrade -y
+apt-get install -y bc binutils binutils-aarch64-linux-gnu binutils-arm-linux-gnueabi bison ccache curl flex expect gcc git gnupg libssl-dev openssl make qemu-system-arm qemu-system-x86
+curl https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+echo "deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch main" | tee -a /etc/apt/sources.list
+apt-get update -qq
+apt-get install -y clang-8 lld-8
--- a/debian-setup.sh
+++ b/debian-setup.sh
@@ -2,9 +2,36 @@
 # Installs all of the necessary packages in a Docker container
 # MUST BE RUN WITH ROOT
 
+# Show all commands and exit upon failure
+set -eux
+
+# Make sure that all packages are up to date
 apt-get update
 apt-get upgrade -y
-apt-get install -y bc binutils binutils-aarch64-linux-gnu binutils-arm-linux-gnueabi bison ccache curl flex expect gcc git gnupg libssl-dev openssl make qemu-system-arm qemu-system-x86
+
+# Install the official Debian packages that we need
+# curl and wget are not installed by default so this
+# is separate from the Clang/lld installation below
+# because we would need to at least install curl to
+# get LLVM's apt key
+apt-get install -y bc \
+                   binutils \
+                   binutils-aarch64-linux-gnu \
+                   binutils-arm-linux-gnueabi \
+                   bison \
+                   ccache \
+                   curl \
+                   flex \
+                   expect \
+                   git \
+                   gnupg \
+                   libssl-dev \
+                   openssl \
+                   make \
+                   qemu-system-arm \
+                   qemu-system-x86
+
+# Install nightly verisons of Clang and lld (apt.llvm.org)
 curl https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
 echo "deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch main" | tee -a /etc/apt/sources.list
 apt-get update -qq
--- a/debian-setup.sh
+++ b/debian-setup.sh
@@ -36,3 +36,9 @@ curl https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
 echo "deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch main" | tee -a /etc/apt/sources.list
 apt-get update -qq
 apt-get install -y clang-8 lld-8
+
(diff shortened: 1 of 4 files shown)


debian-setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Installs all of the necessary packages in a Docker container
+# MUST BE RUN WITH ROOT
```
nickdesaulniers: Does the above `script` section of `.travis.yml` need to run this as root?

debian-setup.sh
```diff
@@ -0,0 +1,11 @@
...
+
+apt-get update
+apt-get upgrade -y
+apt-get install -y bc binutils binutils-aarch64-linux-gnu binutils-arm-linux-gnueabi bison ccache curl flex expect gcc git gnupg libssl-dev openssl make qemu-system-arm qemu-system-x86
```
nickdesaulniers: Do you mind putting these one one line each separated by backslashes? It would make it easier for me to read.

Do we need GCC? 😈

debian-setup.sh
```diff
@@ -0,0 +1,11 @@
...
+curl https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
```
nickdesaulniers: I guess it makes sense to have this separate, but if we did this earlier, we'd only have to update once.  I think it's ok to combine them. What do you think?

nickdesaulniers:
Thanks for this patch! I don't know anything about docker, but it looks like moving to it is relatively painless!

debian-setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Installs all of the necessary packages in a Docker container
+# MUST BE RUN WITH ROOT
+
+apt-get update
+apt-get upgrade -y
```
nickdesaulniers: I'm hesitant to keep the upgrade part; [job 156](https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/156461095) pulled in an update for tzdata for example.  I assume this will just slow down the tests without any meaningful gain?

debian-setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Installs all of the necessary packages in a Docker container
```
nickdesaulniers: `set -eux`

at least `-x` so we can see what's going on in the logs.

debian-setup.sh
nickdesaulniers: I see @shenki put them on separate lines, too. 🥂

debian-setup.sh
nathanchance: `docker`'s default user is root. Kind of a superfluous warning, I can remove it if necessary.

debian-setup.sh
nathanchance: > Do you mind putting these one one line each separated by backslashes? It would make it easier for me to read.

Yes.

> Do we need GCC? 😈

Probably not, will remove :P

debian-setup.sh
nathanchance: I don't think `curl` is installed by default so I'd need to install it for that command then install everything else. I can add a comment explaining why these two things are separate.

debian-setup.sh
nathanchance: As I said to Stefan, the upgrade and initial install of packages is no longer than when we were having Travis do it (in fact, it seems a little quicker). Plus, these Debian images are refreshed every month or so I believe so we're not going to updating that much.

debian-setup.sh
nathanchance: Sure.

debian-setup.sh
nathanchance: Addressed in 18515a3d90b68c06da043f1459e01e45670afe22

debian-setup.sh
nathanchance: Addressed in 18515a3d90b68c06da043f1459e01e45670afe22

debian-setup.sh
nathanchance: Addressed in 18515a3d90b68c06da043f1459e01e45670afe22

debian-setup.sh
nickdesaulniers: Is that important when running docker locally?

debian-setup.sh
nickdesaulniers: But you're upgrading packages we don't care about. We probably should not have been doing it with travis.

Removing it does not break anything, right?

debian-setup.sh
```diff
@@ -0,0 +1,44 @@
...
+# Make sure that all packages are up to date
+apt-get update
+apt-get upgrade -y
+
+# Install the official Debian packages that we need
+# curl and wget are not installed by default so this
+# is separate from the Clang/lld installation below
+# because we would need to at least install curl to
+# get LLVM's apt key
+apt-get install -y bc \
```
nickdesaulniers: would you mind putting bc on its own line? That way if we add something that begins with the letter `a`, we can insert it on its own line without having to update the `apt-get` line.

debian-setup.sh
nathanchance: Don't think so since it should be the same for local builds:

```
$ docker run --rm -ti debian
root@8b6e6bd76cbd:/# 
```

debian-setup.sh
```diff
@@ -0,0 +1,44 @@
...
+                   binutils \
+                   binutils-aarch64-linux-gnu \
+                   binutils-arm-linux-gnueabi \
+                   bison \
+                   ccache \
+                   curl \
+                   flex \
+                   expect \
```
nickdesaulniers: Please sort these (flex vs expect).

If you're using vim, you can do this very quickly:
ctrl+v,j<repeat to highlight lines; maybe a better movement exists>, :sort, enter

debian-setup.sh
nathanchance: Fair point I guess; the only thing I am worried about is a default package requiring an upgrade for an installed package but I guess apt will take care of that (I don't know though, I use Arch Linux personally).

I've always been of the mindset that a system should always be up to date but I guess that doesn't matter for these systems since they're only used for one purpose (plus when Debian's image gets refreshed in Docker, we'll get those updates anyways).

debian-setup.sh
nathanchance: Yes sure.

debian-setup.sh
nathanchance: Oops, yes will do.

debian-setup.sh
nickdesaulniers: > I don't know though, I use Arch Linux personally

Dude, that's like a meme on /r/linux. 🐸 

> I've always been of the mindset that a system should always be up to date

I agree for my personal devices.  For a ci testing platform, I don't think we need as strict a requirement.  And it sounds like the base docker image gets updated automatically after a month or so?  So it's never like the existing packages are ever more than a month out of date.

debian-setup.sh
nickdesaulniers: ok, please remove the comment then.  The caps makes it seems like there's an action item, but I don't see sudo being used in this file, or in the travis.yml, or locally.

debian-setup.sh
nathanchance: > > I don't know though, I use Arch Linux personally
> 
> Dude, that's like a meme on /r/linux. 🐸

Oh believe me, I know :P

> > I've always been of the mindset that a system should always be up to date
> 
> I agree for my personal devices. For a ci testing platform, I don't think we need as strict a requirement. And it sounds like the base docker image gets updated automatically after a month or so? So it's never like the existing packages are ever more than a month out of date.

Hmmm I guess Travis doesn't run `apt-get upgrade`, I thought they did: https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-standard-infrastructure

I am not opposed to removing it if you are worried about stuff breaking (although it is Debian stable haha) and fair point regarding the image updates. I just get triggered when I don't have the latest ;P

Will push the removal here in a second.

nathanchance:
Definitely looks like #23 is addressed as well 😎 (at least for the time being)

https://travis-ci.com/ClangBuiltLinux/continuous-integration/builds/90465336

debian-setup.sh
```diff
@@ -0,0 +1,41 @@
...
+    libssl-dev \
+    make \
+    openssl \
+    qemu-system-arm \
+    qemu-system-x86
+
+# Install nightly verisons of Clang and lld (apt.llvm.org)
+curl https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+echo "deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch main" | tee -a /etc/apt/sources.list
+apt-get update -qq
```
nickdesaulniers: what's up with the `-qq` here but not on L12?

PR #30, opened 2018-11-07
ci: Support linux-next

nathanchance:
It's important to test linux-next as that's where all new development happens.

We don't need to build it every branch push and pull request because it is only updated daily so I only have it building with our daily cron.

I have verified it does work properly on my fork: https://travis-ci.com/nathanchance/continuous-integration/builds/90488778

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: cpp
 matrix:
   include:
+    # linux
     - name: "ARCH=arm"
       env: ARCH=arm
     - name: "ARCH=arm64"
@@ -9,6 +10,19 @@ matrix:
       env: ARCH=arm64 LD=ld.lld-8
     - name: "ARCH=x86_64"
       env: ARCH=x86_64
+    # linux-next
+    - name: "ARCH=arm REPO=linux-next"
+      env: ARCH=arm REPO=linux-next
+      if: type = cron
+    - name: "ARCH=arm64 REPO=linux-next"
+      env: ARCH=arm64 REPO=linux-next
+      if: type = cron
+    - name: "ARCH=arm64 LD=ld.lld REPO=linux-next"
+      env: ARCH=arm64 LD=ld.lld-8 REPO=linux-next
+      if: type = cron
+    - name: "ARCH=x86_64 REPO=linux-next"
+      env: ARCH=x86_64 REPO=linux-next
+      if: type = cron
 compiler: gcc
 os: linux
 dist: trusty
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.tar.gz
 .ccache
 linux/
+linux-next/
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -1,14 +1,26 @@
...
+      env: ARCH=arm REPO=linux
+    - name: "ARCH=arm64 REPO=linux"
+      env: ARCH=arm64 REPO=linux
+    - name: "ARCH=arm64 LD=ld.lld REPO=linux"
+      env: ARCH=arm64 LD=ld.lld-8 REPO=linux
+    - name: "ARCH=x86_64 REPO=linux"
+      env: ARCH=x86_64 REPO=linux
+    - name: "ARCH=arm REPO=linux-next"
+      env: ARCH=arm REPO=linux-next
+      if: type = cron
```
nickdesaulniers: neat, where did you find documentation for `if`? I didn't see any on [matrix](https://docs.travis-ci.com/user/customizing-the-build/) or [cron](https://docs.travis-ci.com/user/cron-jobs/).

nathanchance: Gah meant to add that to the commit message.

https://docs.travis-ci.com/user/conditional-builds-stages-jobs/

https://docs.travis-ci.com/user/conditions-v1

.travis.yml
```diff
@@ -1,14 +1,26 @@
...
-    - name: "ARCH=arm"
-      env: ARCH=arm
-    - name: "ARCH=arm64"
-      env: ARCH=arm64
-    - name: "ARCH=arm64 LD=ld.lld"
-      env: ARCH=arm64 LD=ld.lld-8
-    - name: "ARCH=x86_64"
-      env: ARCH=x86_64
+    - name: "ARCH=arm REPO=linux"
+      env: ARCH=arm REPO=linux
```
nickdesaulniers: Rather than explicitly stating `REPO=linux`, can we just leave it out, since it's the default? Similar to what we do for bfd vs lld for LD?  I think this makes the names a little bit more readable, too.

.travis.yml
nickdesaulniers: Neat; looks good. Just minor bikeshedding about the name and allowing defaults to be implicit. 👽 🌶

.travis.yml
```diff
@@ -1,14 +1,26 @@
...
+    - name: "ARCH=arm64 REPO=linux"
+      env: ARCH=arm64 REPO=linux
```
nickdesaulniers: I feel like with these building blocks (conditional on cron) we could make non-lld arm64 builds cron only.

Idea being that on our free tier, we only have access to three concurrent builds. So to to have fast presubmit feedback, we want to minimize our PR builds to multiples of three (preferably only three).

Is that a bad idea? 🎲 📍

nathanchance:
Yes fair point, done: ee9c829c3d8527fd3afc93ab8c165f81db108658

Turns out this is also beneficial because adding a new variable creates a
new cache...

nathanchance:
I wouldn't be opposed to this. I can either do that in this pull or another
one, up to you.

.travis.yml
nathanchance: (reposting since replying via email doesn't keep comments inline)

Yes fair point, done: ee9c829c3d8527fd3afc93ab8c165f81db108658

Turns out this might be beneficial since adding a new variable appears to create a
new cache...

.travis.yml
nathanchance: I wouldn't be opposed to this. I can either do that in this pull or another one, up to you.

PR #31, opened 2018-11-07
.travis.yml: Demote arm64 ld.bfd build to cron only

nathanchance:
Nick suggested moving this build to cron only in #30 to speed up pull
request validation because at our free tier, we only have access to
three concurrent jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,15 @@ matrix:
     # linux
     - name: "ARCH=arm"
       env: ARCH=arm
-    - name: "ARCH=arm64"
-      env: ARCH=arm64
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
Looks like LLD is stuck linking the final vmlinux image?
```
  LD      vmlinux
```

nathanchance:
Yeah I restarted the build to see if it's just Travis acting up or a real issue (although I doubt it since #30 was fine lol).

nickdesaulniers:
weird; same version of clang, so it's not like LLD updated between runs. Stupid computers 🦆

PR #32, opened 2018-11-07
driver.sh: Build with Guenter Roeck's extra config options

nathanchance:
While we won't use every single configuration option selected here
during our boot tests like Guenter does, it hould give us a little more
coverage on warnings. The self tests run during the QEMU boot so we can
make sure that Clang isn't breaking anything fundamental.


nathanchance:
@groeck if you feel something should be removed or added, let me know.

This is the last remaining pending addition that I have planned for this repo, I think we should be in a really good state after this :)

driver.sh
```diff
@@ -129,6 +129,78 @@ build_linux() {
...
+CONFIG_USB_UAS=y
+CONFIG_VIRTIO=y
+CONFIG_VIRTIO_PCI=y
+CONFIG_VIRTIO_PCI_LEGACY=y
+CONFIG_VIRTIO_BALLOON=y
+CONFIG_VIRTIO_MMIO=y
+CONFIG_BLK_MQ_VIRTIO=y
+CONFIG_VIRTIO_BLK=y
+CONFIG_VIRTIO_BLK_SCSI=y
+CONFIG_SCSI_VIRTIO=y" >> .config
```
nickdesaulniers: Put this is a different file checked in (`git add ...`), then copy it over to `.config`.

nickdesaulniers: as in, this should be a separate config file, rather than a big string literal in a shell script. Then the shell script can simply do `cp groeck_config .config` for a given target.

nathanchance: These configs are meant to supplement a defconfig, not be a complete config fragment but I agree that they should be in a separate file then concatenated onto `.config`.

nathanchance:
To be fair regarding the build times, the cache is probably cold now since the config options have been changed. I'm sure that the build wouldn't be as long in subsequent runs.

groeck:
FWIW, I am fine with the changes. They should not result in significant build time changes. Boot time will be a bit higher, though, due to the executed self tests.

nathanchance:
> Good point. Maybe rekick it a few times so we get a better sense of the hot cache performance?

Yeah once #33 is merged, I'll run this a couple times to make sure the cache isn't impacted.

> I still think this could be an additional target in the testing matrix, rather than polluting the defconfig builds. 🤷‍♂️

I'm just worried about our build matrix getting too gnarly. We still have PowerPC and ARMv6 lined up and talks about adding `tiny.config` and `all{yes,mod}config` too. These seem like sane config options to test alongside `defconfig` but I guess I can concede and we can have two config options.

> FWIW, I am fine with the changes. They should not result in significant build time changes. Boot time will be a bit higher, though, due to the executed self tests.

Thanks for the reply. I've noticed the higher boot times but haven't seen anything go over 30s or so, which seems reasonable to me.

nickdesaulniers:
> Boot time will be a bit higher, though, due to the executed self tests.

So the boot tests are run automatically? Will they be run before our init script powers down the machine?

nathanchance:
Yup: https://travis-ci.com/nathanchance/continuous-integration/jobs/156862459#L7138

driver.sh
```diff
@@ -129,6 +129,14 @@ build_linux() {
   # is good about figuring out what needs to be rebuilt
   [[ -n "${cleanup:-}" ]] && mako_reactor mrproper
   mako_reactor ${config}
+  # If we're using a defconfig, enable some more common config options
+  # like debugging, selftests, and common drivers
+  if [[ ${config} =~ defconfig ]]; then
```
nickdesaulniers: was `=~` intentional (regex) vs just `==` for string equality?

nathanchance: Yes because of `multi_v7_defconfig` (although that's the target of `ARCH=arm make defconfig` anyways so maybe that should be updated)

driver.sh
```diff
@@ -129,6 +129,14 @@ build_linux() {
...
+    cat ../configs/common.config >> .config
+    # Some torture test configs cause issues on x86_64
+    [[ $ARCH != "x86_64" ]] && cat ../common/tt.config >> .config
+  fi
+  mako_reactor olddefconfig
```
nickdesaulniers: neat, I did not know about `olddefconfig`.

nickdesaulniers: does it always print these warnings?
```
.config:4409:warning: override: reassigning to symbol EXPERT
```

nathanchance: I believe so given that the options are being overridden. I can hide that behind a `>/dev/null` if necessary.

groeck:
> > Boot time will be a bit higher, though, due to the executed self tests.
> 
> So the boot tests are run automatically? Will they be run before our init script powers down the machine?

Correct, tests enabled in the configuration files (such as spinlock torture tests or the locking API tests suite) run automatically, typically before handover to userspace. Some keep running, though.

driver.sh
nickdesaulniers: Yes please.

groeck:
On Wed, Nov 07, 2018 at 08:59:43AM -0800, Nick Desaulniers wrote:
> @groeck 
> 
> In our logs, I see:
> ```
> [    0.005861] | Locking API testsuite:
> ...
> [    0.123536]   mixed read-lock/lock-write ABBA:             |FAILED|             |  ok  |
> ...
> [    0.441747] Good, all 261 testcases passed! |
> ```
> 
> Is that an expected failure? Seems odd to print `FAILED` and yet `all 261 testcases passed!`.
> 
Oddly enough yes. The failures depend on the architecture.
I never checked the actual code trying to find out how the test 
determines pass/fail expectations per architecture.

> Also, do you know: when a test suite fails, does it cause a panic?  Cause we'll need to verify that a panic in the kernel causes our integration test to fail.
> 
Unfortunately, you'll have to parse the output. The only time where I have
seen a crash as result of a failed test was a bad error message. On top of
that, the output differs from test to test. This wasn't designed with
automation in mind.

It might make sense to start a discussion if the error messages and their
handling can be unified for automation. Note that some tests are known to
fail, such as some locking tests on arm64 with 8+ CPUs. A panic on failure
might thus be counter-productive. A WARN on a test failure might be useful,
though.

Guenter

PR #33, opened 2018-11-07
debian-setup: quiet down setup, and zero stats

nickdesaulniers:
The logs are full of noise about packages being installed. This is
noise.

Printing the stats before and after makes it hard to tell the difference
without doing math.  It's clearer to zero stats before. Then the stats
after are just for this build.  We don't care about the cache stats pre
run.  Clearing the stats DOES NOT clear the cache itself or reset its
size.

https://ccache.samba.org/manual/latest.html#_options

### DIFF
--- a/debian-setup.sh
+++ b/debian-setup.sh
@@ -9,8 +9,8 @@ set -eux
 # is separate from the Clang/lld installation below
 # because we would need to at least install curl to
 # get LLVM's apt key
-apt-get update
-apt-get install -y \
+apt-get update -qq
+apt-get install -y -qq \
     bc \
     binutils \
     binutils-aarch64-linux-gnu \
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
> I assume the additonal slashes at the end of the apt-get install blocks were added so that adding additional packages is a one line change, rather than two?

Yep; I probably don't need to be zealous about protecting `git blame` for such a small project.  But sometimes when you want to know *why* a statement was added; `git blame` can you point you to a commit that can often answer *why*.

nickdesaulniers:
Saves ~1500 lines from the beginning of the logs. :open_mouth:

nickdesaulniers:
eh, looks like stdout or stderr needs to be redirected to /dev/null.

nathanchance:
It might need to be `apt-get -qq`: https://askubuntu.com/questions/258219/how-do-i-make-apt-get-install-less-noisy

Could just make it an wrapper function if that's the case:

```
function apt-get() {
  command apt-get -qq "${@}"
}
```

nickdesaulniers:
haha ok good find

nathanchance:
Just tried it locally, doesn't appear to work.

nickdesaulniers:
heh, the comments in that link mention similar issues.  Logging is probably from dpkg.

debian-setup.sh
```diff
@@ -27,6 +27,7 @@ apt-get install -y -qq \
     openssl \
     qemu-system-arm \
     qemu-system-x86 \
+    &>/dev/null
```
nathanchance: Hmmm do we want to direct both stderr AND stdout to `/dev/null`? If there is an error during install, we'd probably want to see it.

nickdesaulniers: done

nickdesaulniers:
> cache hit rate                    100.00 %

:dollar: :dollar: :heavy_dollar_sign:

PR #36, opened 2018-11-09
Switch to Debian sid and Qemu 3.0

shenki:
These changes are in preparation for enabling ppc64le.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ cache:
 services:
   - docker
 script:
-  - docker create -v ${TRAVIS_BUILD_DIR}:/travis --name travis debian
-  - docker run --env ARCH=${ARCH} --env LD=${LD} --rm --volumes-from travis debian /bin/bash -c 'cd /travis && ./debian-setup.sh && ./driver.sh && ccache -s'
+  - docker create -v ${TRAVIS_BUILD_DIR}:/travis --name travis debian:sid
+  - docker run --env ARCH=${ARCH} --env LD=${LD} --rm --volumes-from travis debian:sid /bin/bash -c 'cd /travis && ./debian-setup.sh && ./driver.sh && ccache -s'
 after_script:
   - sleep 1
--- a/debian-setup.sh
+++ b/debian-setup.sh
@@ -25,6 +25,14 @@ apt-get install -y -qq \
     libssl-dev \
     make \
     openssl \
+    >/dev/null
+
+# Install Qemu 3.0 from shenki's ppa. Debian nor Ubuntu package 3.0 yet, which
+# is required for ppc64le powernv support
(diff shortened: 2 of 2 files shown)


debian-setup.sh
```diff
@@ -25,10 +25,18 @@ apt-get install -y -qq \
...
 
+
+# Install Qemu 3.0 from shenki's ppa. Debian nor Ubuntu package 3.0 yet, which
+# is required for ppc64le powernv support
+apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E007EC6A
+echo "deb http://ppa.launchpad.net/shenki/ppa/ubuntu bionic main" | tee -a /etc/apt/sources.list
+apt-get update -qq
+apt-get install -y -qq \
+    qemu-system-arm \
+    qemu-system-x86
```
nickdesaulniers: please redirect stdout to /dev/null as in L28 and L47. 😯

.travis.yml
```diff
@@ -35,7 +35,7 @@ cache:
 services:
   - docker
 script:
-  - docker create -v ${TRAVIS_BUILD_DIR}:/travis --name travis debian
-  - docker run --env ARCH=${ARCH} --env LD=${LD} --rm --volumes-from travis debian /bin/bash -c 'cd /travis && ./debian-setup.sh && ./driver.sh && ccache -s'
+  - docker create -v ${TRAVIS_BUILD_DIR}:/travis --name travis debian:sid
+  - docker run --env ARCH=${ARCH} --env LD=${LD} --rm --volumes-from travis debian:sid /bin/bash -c 'cd /travis && ./debian-setup.sh && ./driver.sh && ccache -s'
```
nickdesaulniers: What's the default if you don't specify?

nickdesaulniers:
Cool! Thanks for the patches! 🏎

.travis.yml
shenki: Stable, I think:
```
$ docker run -it debian cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
```

debian-setup.sh
shenki: :+1:

.travis.yml
nathanchance: Yeah, it's just a Stretch image by default.

debian-setup.sh
```diff
@@ -25,10 +25,18 @@ apt-get install -y -qq \
     libssl-dev \
     make \
     openssl \
-    qemu-system-arm \
-    qemu-system-x86 \
     >/dev/null
 
+
```
nathanchance: I think this extra line should go (or two spaces should be added in other places).

PR #37, opened 2018-11-09
Support ppc64le

shenki:


.travis.yml
```diff
@@ -8,6 +8,8 @@ matrix:
       env: ARCH=arm64 LD=ld.lld-8
     - name: "ARCH=x86_64"
       env: ARCH=x86_64
+    - name: "ARCH=ppc64le"
+      env: ARCH=ppc64le
```
nickdesaulniers: Additionally, let's add an additional `REPO=linux-next` target `if: type = cron` below.

driver.sh
```diff
@@ -70,6 +70,17 @@ setup_variables() {
       qemu_cmdline=( -drive "file=images/x86_64/rootfs.ext4,format=raw,if=ide"
                      -append "console=ttyS0 root=/dev/sda" ) ;;
 
+    "ppc64le")
+      config=powernv_defconfig
+      image_name=zImage.epapr
+      qemu="qemu-system-ppc64"
+      qemu_cmdline=( -M powernv -m 3G -nographic
```
nickdesaulniers: Does the `-m 3G` here conflict with the one one L162? (how much memory is allotted to the VM if you specify `-m` twice?)  Can you boot in `-m 512`?

driver.sh
```diff
@@ -70,6 +70,17 @@ setup_variables() {
...
+                     -device ipmi-bmc-sim,id=bmc0 -device isa-ipmi-bt,bmc=bmc0,irq=10
```
nickdesaulniers: Would you mind putting these on separate lines if the current line L78 runs over 80 chars? (I can't tell from GH UI).

nickdesaulniers:
I'm super happy to see PPC support 🔋 !! Thanks @shenki

nickdesaulniers:
oops, didn't mean to click approve. Please make the requested changes before submitting.  Seems like I can't redact a review in GH UI?

nickdesaulniers:
@shenki would you please add a script similar to how we do manual buildroot rootfs builds, that way other people may update the skiboot if ever needed?

driver.sh
nickdesaulniers: If not, I suggest adding `-m 512` to the other case statements, and removing the default from L162.

.travis.yml
shenki: Thanks. I also sorted the entries.

driver.sh
shenki: The entry that comes last takes precedence. This is why I moved the `${qemu_cmdline[@]}` to after `-m 512` in `boot_qemu()`.

We can't boot with less due to some limitations (I think in Qemu? I'm chasing it up with some people who know the details).

I'll make it explicit for all entries.

shenki:
> @shenki would you please add a script similar to how we do manual buildroot/rebuild.sh rootfs builds but for the skiboot, that way other people may update the skiboot if ever needed?

Sure. I've opened https://github.com/ClangBuiltLinux/continuous-integration/issues/38 as I'd like to try pulling it in through the distro instead of having another binary in the tree.

driver.sh
groeck: 2G works for me when booting a powernv image.

shenki:
For reference, to build we do this:
```
apt install gcc-powerpc64-linux-gnu binutils-powerpc64-linux-gnu
git clone https://github.com/open-power/skiboot
cd skiboot
CROSS=powerpc64-linux-gnu- make
```

driver.sh
groeck: qemu-system-ppc64 -M powernv -cpu POWER8 -m 2G -kernel arch/powerpc/boot/zImage.epapr -initrd /var/cache/buildbot/ppc64/rootfs-el.cpio -nographic -vga none -monitor null -no-reboot --append 'rdinit=/sbin/init panic=-1 console=tty console=hvc0'

1G crashes with a NULL pointer access in unpack_to_rootfs().

shenki:
@groeck The Debian packaging process strips the binaries out, so we don't have it there. They then package up the firmware projects (built from source) and have the Qemu package depend on the firmware packages.

There's no benefit to building our own at this stage. We do have WIP Qemu patches in github.com/legoater/qemu that model more of the Power9 system, and require some skiboot patches, but I think that's out of scope for this project. We will benefit from it in the Qemu 3.2 timeline.

.travis.yml
nickdesaulniers: did you push that change?

driver.sh
```diff
@@ -19,7 +19,7 @@ setup_variables() {
         echo "   They can be invoked either via 'export VAR=<value>; ./driver.sh' OR 'VAR=value ./driver.sh'"
         echo
         echo "   ARCH:"
-        echo "       If no ARCH value is specified, arm64 is the default. Currently, arm, arm64, and x86_64 are supported."
+        echo "       If no ARCH value is specified, arm64 is the default. Currently, arm, arm64, x86_64 and ppc64le are supported."
```
nickdesaulniers: sorry, this probably needs a rebase now.

driver.sh
shenki: @groeck Thanks. I will sort out that crash and we can look at reducing the amount of ram in the future.

shenki:
Nice. I'll send a v2 of my patch when I get a moment.

PR #39, opened 2018-11-10
driver: move usage to separate file, displayed with cat

nickdesaulniers:


usage.txt
```diff
@@ -0,0 +1,25 @@
...
+Environment variables:
+  The script can take into account specific environment variables, mostly used
+  with Travis.  They can be invoked either via 'export VAR=<value>;
+  ./driver.sh' OR 'VAR=value ./driver.sh'
+
+  ARCH:
+      If no ARCH value is specified, arm64 is the default. Currently, arm,
+      arm64, and x86_64 are supported.
+  LD:
+      If no LD value is specified, \${CROSS_COMPILE}-ld is used. arm64 only.
```
groeck: Is the backslash here still needed ?

nickdesaulniers: done

Issue #35, opened 2018-11-08
Add LTS branches back to 4.4


nickdesaulniers:
We first shipped a clang built 4.4 kernel on Pixel 2 (though we did have a patchset for 3.18 working on Pixel 1).  Now that we're testing torvalds/linux and -next, it would be good to add support for LTS branches (but only back to 4.4).

I'm thinking of moving the switch statements to some kind of hierarchical config or data structure.  I think it that way, it will make adding more targets easier in the future.

@tom-gall sent me some configs based off these LTS branches, but I think having the defconfigs for these LTS branches being tested is a good first step towards that goal.

nathanchance:
Given that the Clang patchset is out of tree for 4.4 and 4.9, how did you want to maintain that? Patch files or a whole repo?

nickdesaulniers:
I spoke with @gregkh about this in person.  IIRC, he said "if you need anything for building LTS in Clang just send it to me."  We want them for Android common kernels anyways, so it's simpler to land things upstream from their in LTS and let them flow down.

nathanchance:
I figured, @mkaehlcke had asked if Clang support could be added directly to LTS earlier this year: https://android-review.googlesource.com/c/kernel/common/+/654139

The Chromium backports were used for kernel/common (as I am sure you know 😛) and should have the upstream SHAs in the messages for inclusion in -stable.

4.4: https://chromium.googlesource.com/chromiumos/third_party/kernel/+log/refs/sandbox/mka/llvm/v4.4

4.9: https://chromium.googlesource.com/chromiumos/third_party/kernel/+log/refs/sandbox/mka/llvm/v4.9

The Pixel 2 kernel shipped [the pre-polished patches](https://android.googlesource.com/kernel/msm/+log/84f212d45dd3a80df381017e6e837592caf744e6) it seems but all of those should have found their way upstream. I also did my own backports for 3.18 two separate times (purely for reference of commits, not to send to stable), including the core commits + some warning fixes:

https://github.com/nathanchance/marlin/commits/76449b487e33ff90720258f39d347ba02a973e78

https://github.com/nathanchance/android_kernel_google_dragon/commits/895b05a174c77afefe2b1de301c698f35f80a8b8

nickdesaulniers:
[...]

PR #40, opened 2018-11-10
add arm64 LTS targets

nickdesaulniers:
this obviously conflicts with #39 , so will rebase on top.

@ajs1984 and I discussed today seeing what was broken in these branches in stable, and sending the relevant backports to @gregkh rather than landing them in Android common kernels, since they merge down from LTS anyways.

We should move these to cron jobs (post submit), but doing so now would make it so that these new targets aren't tested.  Once we get a run, I'll either make the cron, or comment them out (if they're super broken) with a patch on top of this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,20 @@ matrix:
     - name: "ARCH=x86_64 REPO=linux-next"
       env: ARCH=x86_64 REPO=linux-next
       if: type = cron
+    # stable
+    # TODO: re-enable once green
+    #- name: "ARCH=arm64 REPO=4.19"
+      #env: ARCH=arm64 REPO=4.19
+      #if: type = cron
+    - name: "ARCH=arm64 REPO=4.14"
+      env: ARCH=arm64 REPO=4.14
+      if: type = cron
+    #- name: "ARCH=arm64 REPO=4.9"
+      #env: ARCH=arm64 REPO=4.9
+      #if: type = cron
+    #- name: "ARCH=arm64 REPO=4.4"
+      #env: ARCH=arm64 REPO=4.4
+      #if: type = cron
 compiler: gcc
 os: linux
 dist: trusty
--- a/driver.sh
+++ b/driver.sh
@@ -69,6 +69,7 @@ setup_variables() {
       echo "Unknown ARCH specified!"
       exit 1 ;;
   esac
+  export ARCH=${ARCH}
 
   # torvalds/linux is the default repo if nothing is specified
   case ${REPO:=linux} in
--- a/driver.sh
+++ b/driver.sh
@@ -73,9 +73,16 @@ setup_variables() {
 
   # torvalds/linux is the default repo if nothing is specified
   case ${REPO:=linux} in
-    "linux") owner=torvalds ;;
-    "linux-next") owner=next ;;
+    "linux")
+      owner=torvalds ;;
+    "linux-next")
+      owner=next
+      tree=linux-next ;;
+    "4.4"|"4.9"|"4.14"|"4.19")
+      owner=stable
+      branch=linux-${REPO}.y ;;
   esac
+  url=git://git.kernel.org/pub/scm/linux/kernel/git/${owner}/${tree:=linux}.git
 }
 
 check_dependencies() {
@@ -106,13 +113,13 @@ mako_reactor() {
 build_linux() {
   CC="$(command -v ccache) $(command -v clang-8)"
 
-  if [[ -d ${REPO} ]]; then
-    cd ${REPO}
-    git fetch --depth=1 origin master
-    git reset --hard origin/master
+  if [[ -d ${tree} ]]; then
+    cd ${tree}
+    git fetch --depth=1 ${url} ${branch:=master}
+    git reset --hard FETCH_HEAD
   else
-    git clone --depth=1 git://git.kernel.org/pub/scm/linux/kernel/git/${owner}/${REPO}.git
-    cd ${REPO}
+    git clone --depth=1 -b ${branch:=master} --single-branch ${url}
+    cd ${tree}
   fi
 
   git show -s | cat
@@ -138,7 +145,7 @@ build_linux() {
 }
 
 boot_qemu() {
-  local kernel_image=${REPO}/arch/${ARCH}/boot/${image_name}
+  local kernel_image=${tree}/arch/${ARCH}/boot/${image_name}
   # for the rest of the script, particularly qemu
   set -e
   test -e ${kernel_image}
--- a/usage.txt
+++ b/usage.txt
@@ -13,7 +13,13 @@ Environment variables:
   LD:
       If no LD value is specified, ${CROSS_COMPILE}-ld is used. arm64 only.
   REPO:
-      linux (default) or linux-next, to specify which tree to clone and build.
+      Nicknames for trees:
+        linux (default)
+        linux-next
+        4.19
+        4.14
+        4.9
+        4.4
 
 Optional parameters:
   -c | --clean:


nickdesaulniers:
Fixes #35

.travis.yml
```diff
@@ -25,6 +25,16 @@ matrix:
     - name: "ARCH=x86_64 REPO=linux-next"
       env: ARCH=x86_64 REPO=linux-next
       if: type = cron
+    # stable
+    # TODO: make cron only
+    - name: "ARCH=arm64 REPO=4.18"
+      env: ARCH=arm64 REPO=4.18
```
nickdesaulniers: maybe this needs to be `REPO="4.18"`? 🤷‍♂️

nickdesaulniers:
Not sure if 4.19 is also an LTS branch? Seems kind of close to 4.18, though?

.travis.yml
```diff
@@ -25,6 +25,16 @@ matrix:
     - name: "ARCH=x86_64 REPO=linux-next"
       env: ARCH=x86_64 REPO=linux-next
       if: type = cron
+    # stable
+    # TODO: make cron only
+    - name: "ARCH=arm64 REPO=4.18"
+      env: ARCH=arm64 REPO="4.18"
```
nickdesaulniers: strange, the logs show the default value being used below, as if this is not set.

nickdesaulniers:
needs a rebase on #41 too.

nickdesaulniers:
@groeck do you know who maintains: https://www.kernel.org/ ? It says 4.18 AND 4.19 are stable.

Thanks for the the clarification though, I'll fix that in my patch.

groeck:
> @groeck do you know who maintains: https://www.kernel.org/ ? It says 4.18 AND 4.19 are stable.
> 

Misunderstanding. "stable" does not mean "longterm". I assume they'll update the 4.19 tag to "longterm" in due time, maybe after 4.18 is marked EOL. If you search for "Linux LTS 4.19" you'll find references to 4.19 being the next longterm stable release.

driver.sh
```diff
@@ -75,12 +81,36 @@ setup_variables() {
...
+    "4-9")
+      owner=stable
+      tree=linux
+      branch=linux-4.9.y ;;
+    "4-4")
+      owner=stable
+      tree=linux
+      branch=linux-4.4.y ;;
   esac
+  url=git://git.kernel.org/pub/scm/linux/kernel/git/${owner}/${tree}.git
```
nathanchance: We could cut down on a ton of `tree` definitions by making this `${tree:=linux}`

.travis.yml
```diff
@@ -25,6 +25,16 @@ matrix:
     - name: "ARCH=x86_64 REPO=linux-next"
       env: ARCH=x86_64 REPO=linux-next
       if: type = cron
+    # stable
+    # TODO: make cron only
+    - name: "ARCH=arm64 REPO=4-18"
+      env: ARCH=arm64 REPO=4-18
+    - name: "ARCH=arm64 REPO=4-14"
+      env: ARCH=arm64 REPO=4-14
+    - name: "ARCH=arm64 REPO=4-9"
```
nathanchance: Should 4.4 and 4.9 be added just yet since support is not upstream yet?

driver.sh
```diff
@@ -137,7 +167,7 @@ build_linux() {
     [[ $ARCH != "x86_64" ]] && cat ../configs/tt.config >> .config
     mako_reactor olddefconfig &>/dev/null
   fi
-  mako_reactor ${image_name}
+  mako_reactor
```
nathanchance: Did you mean to remove this?

driver.sh
```diff
@@ -75,12 +81,36 @@ setup_variables() {
...
+      branch=linux-4.18.y ;;
+    "4-14")
+      owner=stable
+      tree=linux
+      branch=linux-4.14.y ;;
+    "4-9")
+      owner=stable
+      tree=linux
+      branch=linux-4.9.y ;;
+    "4-4")
```
nathanchance: Hopefully once #42 is merged, all of these could be combined into:

```
"4.4"|"4.9"|"4.14"|"4.18"|"4.19")
  owner=stable
  branch=linux-${REPO}.y ;;
```

driver.sh
nickdesaulniers: IIRC, 4.14 locally failed (no such target).  Let me triple check.   Thanks for the reference to the commit; I didn't realize we were saving build time by not building default modules.

.travis.yml
nickdesaulniers: I guess it's a question of having things wired up, even if they're red on day one.  Depends on what we want to do if/when a target goes red; do we want to disable it, or let every cron be red until a fix lands?

I'd prefer to have some automated testing so we can collect warnings from the logs. Maybe we can settle the question about regressions later.

driver.sh
nickdesaulniers: It's a good recommendation for minimizing duplication between existing targets.

driver.sh
nickdesaulniers: nvm, seems to work.

.travis.yml
nathanchance: I'd argue we should keep everything green. Since 4.14 and 4.19 _should_ work fine and the case block below is super easy to extend with new version (and 4.4 and 4.9 can even remain), I suggest we keep 4.4 and 4.9 off of the matrix until the core Clang support is added to those branches upstream. Alternatively, we could create a `linux-stable` repo on our organization (or branches on the main `linux` repo) and have the backports there until integration upstream just for early testing.

nickdesaulniers: Ok, I've updated the patch set.  I want at least one run presubmit, just so we have logs we can point to from our issue tracker or whatever to keep track of what needs to get backported.  Will comment out the configs once we get a run.

nickdesaulniers:
I need to fix the variable `kernel_image`, it's not correct for the new `REPO`s.

> + test -e 4.14/arch/arm64/boot/Image.gz

nathanchance:
Ah yes should be `tree`, thank goodness for presubmit.

nathanchance:
4.19 needs https://github.com/ClangBuiltLinux/linux/commit/2a6c7c367de82951c98a290a21156770f6f82c84

4.4/4.9 need the series mentioned in #35.

Hopefully that patch makes it into 4.19.2, Greg seems to be churning through his backlog pretty quickly.

Since it is only one patch that should be needed, we could add a `patches/${REPO}` folder to apply it in the meantime: https://github.com/nathanchance/continuous-integration/commit/afc7076c06573e9ce75311e821ede1b59b2f4a31

It'll be tested here, I have to run off to work: https://travis-ci.com/nathanchance/continuous-integration/jobs/157776786

nickdesaulniers:
I just got an email that @gregkh added the -fcall-used-x0 to stable. (an hour after your post @nathanchance )

nathanchance:
Yeah ah the irony lol

nickdesaulniers:
ok, now that we have a presubmit run (build 225), let's commit just 4.14 support.

nickdesaulniers:
> Makefile:966: *** "Cannot generate ORC metadata for CONFIG_UNWINDER_ORC=y, please install libelf-dev, libelf-devel or elfutils-libelf-devel".  Stop.

Did the ORC unwinder just become default on?

nathanchance:
No, my '-pipe' patch got merged so the `git apply` command failed. I'm actually testing dropping the patch right now: https://github.com/nathanchance/continuous-integration/commit/7b32d0959329ded5f429d1ad7a322c2f443bae30

nickdesaulniers:
`set -e` would have stopped the build at the error. :stuck_out_tongue_closed_eyes:

nathanchance:
It might be worth setting that globally if no command is meant to fail.

PR #46, opened 2018-11-12
driver: small cleanup

nickdesaulniers:
prefer:
if x:
  ...
else:
  ...

to:
if !x:
  ...
else:
  ...


nathanchance:
This is very clearly an equivalent before and after and we'd need a rebase + another Travis run to verify that everything is fine. I'm merging this without the presubmit.

PR #47, opened 2018-11-12
patches: x86_64: Remove '-pipe' patch

nathanchance:
It's been merged: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=437e88ab8f9e2ad90576ab74c4cf8f527bbf51cd


nickdesaulniers:
thanks for dropping this fix now that it has landed upstream!

PR #48, opened 2018-11-12
debian-setup: Install libelf-dev

MSF-Jarvis:
The job failure at [1] appears to have been due to the libelf-dev package being missing. It is available in the sid repositories [2] so we can install it directly.

[1]: https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/157821684#L576
[2]: https://packages.debian.org/sid/libelf-dev

### DIFF
--- a/debian-setup.sh
+++ b/debian-setup.sh
@@ -23,6 +23,7 @@ apt-get install -y -qq \
     flex \
     git \
     gnupg \
+    libelf-dev \
     libssl-dev \
     make \
     openssl \


nathanchance:
That fail is fixed in #47 (because the out of tree patch that removes that failure message didn't get applied) but this is probably worth adding, I'll defer to Nick for further comment.

debian-setup.sh
```diff
@@ -24,6 +24,7 @@ apt-get install -y -qq \
     git \
     gnupg \
     libssl-dev \
+    libelf-dev \
```
nathanchance: Although this should be alphabetized

nathanchance:
Alright, this pull is good and is needed (verified with [commit](https://github.com/nathanchance/continuous-integration/commit/8796a82f376b7509c062b8c286378c495d8765a4) + [Travis build](https://travis-ci.com/nathanchance/continuous-integration/builds/91060888), which I will push after this is merged) but the build will not pass until #47 is merged.

nickdesaulniers:
thanks @MSF-Jarvis !!!

PR #53, opened 2018-11-12
Some Travis cleanup/simplification

nathanchance:


nathanchance:
I'd say if we need to add any more commands, that's what we should do.

PR #54, opened 2018-11-12
driver: Build with CONFIG_DEBUG_SECTION_MISMATCH

nathanchance:
I should have fixed all of these mismatches but we should be building
with this to make sure we catch them and file them appropriately.


driver.sh
```diff
@@ -131,7 +131,7 @@ build_linux() {
     [[ $ARCH != "x86_64" ]] && cat ../configs/tt.config >> .config
     mako_reactor olddefconfig &>/dev/null
   fi
-  mako_reactor ${image_name}
+  mako_reactor CONFIG_DEBUG_SECTION_MISMATCH=y ${image_name}
```
nickdesaulniers: this works? :open_mouth: how about `scripts/config.sh -e ...`? (preferred)

nathanchance: This is how the build system tells us to do it:

```
WARNING: modpost: Found 6 section mismatch(es).
To see full details build your kernel with:
'make CONFIG_DEBUG_SECTION_MISMATCH=y'
```

nathanchance: Done.

driver.sh
```diff
@@ -131,6 +131,10 @@ build_linux() {
     [[ $ARCH != "x86_64" ]] && cat ../configs/tt.config >> .config
     mako_reactor olddefconfig &>/dev/null
   fi
+  # Make sure we build with CONFIG_DEBUG_SECTION_MISMATCH so that the
+  # full warning gets printed and we can file and fix it properly.
+  ./scripts/config -e DEBUG_SECTION_MISMATCH
+  mako_reactor olddefconfig &>/dev/null
```
nickdesaulniers: you can remove L132 then, as it's the same as L137 unconditionally now.

Also, why do you do:
1. make defconfig
2. scripts/config -e ...
3. make olddefconfig

I would think 3 is unnecessary?

nathanchance:
Looks like the olddefconfig is still needed for older builds: https://travis-ci.com/nathanchance/continuous-integration/jobs/158108044

I've made your suggested change though.

PR #55, opened 2018-11-12
driver: Call 'set -e' globally

nathanchance:
We expect all of the commands to succeed and we want the script to error
if that is not the case.

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -u
+set -eu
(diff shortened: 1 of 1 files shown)


nathanchance:
Doesn't appear so otherwise ARM/PPC would have failed (`[[ -d ${patches_folder} ]] && git apply -3 "${patches_folder}"/*.patch`): https://travis-ci.com/nathanchance/continuous-integration/builds/91134383

PR #56, opened 2018-11-13
buildroot: x86_64: Use 4.4 kernel headers for toolchain

nathanchance:
The default is 4.17, which causes issues when we try to boot kernels
that are older than that.

https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/158053429


nathanchance:
It wasn't needed for arm64 because that rootfs was build with a Linaro toolchain which should use older headers. However, I'm going to switch to the buildroot toolchain to avoid the issue noted in 1f773ebd677460ca5a9f0620374c1fe3dac3fae1. It will be necessary for arm and ppc, I'm going to test arm with the small backport I did for my Raspberry Pi here shortly.

PR #44, opened 2018-11-12
X86 LTS

nickdesaulniers:
branched off of #40 . Let's see what's green for x86.

PR #57, opened 2018-11-13
Build all rootfs images with a toolchain with 4.4 headers + uClibc-ng

nathanchance:
Follow up to #56, we will need this to be able to boot kernels older than 4.17 (buildroot's default). Additionally, switching to uClibc-ng allows us to save quite a bit of space with the cpio images and might even result in a faster booting kernel image.

I tested arm{32,64}. I didn't test PowerPC since I am not sure what is needed in terms of backports (if that is even feasible).


buildroot/arm.config
```diff
@@ -1,5 +1,9 @@
 BR2_arm=y
 BR2_OPTIMIZE_3=y
+BR2_TOOLCHAIN_BUILDROOT_GLIBC=y
```
shenki: Why are we forcing glibc here? (I assume that's the cause of the +1.93MB to the cpio)

nathanchance: No particular reason, was just making it match everything else.

I suppose if we save space by switching to another libc, that's probably worth doing.

shenki:
PowerPC should work with older kernels.

Out of interest, what fails when you boot an image with the newer headers on an old kernel?

nathanchance:
https://github.com/ClangBuiltLinux/continuous-integration/pull/44#issuecomment-438096328

buildroot/arm.config
nathanchance: Yeah using uClibc-ng on arm64 shrinks the cpio image by 35%: https://github.com/nathanchance/continuous-integration/commit/4fe34b885404493180a1ac565cadd86e0eeb963e

I'll push that change + reverting the arm32 image back to it as soon as Travis verifies everything works properly.

shenki:
> PowerPC should work with older kernels.

This comment brought to you pre-coffee.

Of course 4.4 doesn't have all of the clang fixes, so it won't build unless someone collects the patches and backports them.

buildroot/x86_64.config
```diff
@@ -1,7 +1,6 @@
 BR2_x86_64=y
 BR2_x86_corei7=y
 BR2_OPTIMIZE_3=y
-BR2_TOOLCHAIN_BUILDROOT_GLIBC=y
```
nickdesaulniers: So this was only set for x86?

buildroot/arm64.config
```diff
@@ -1,7 +1,9 @@
 BR2_aarch64=y
 BR2_cortex_a57=y
 BR2_OPTIMIZE_3=y
-BR2_TOOLCHAIN_EXTERNAL=y
+BR2_KERNEL_HEADERS_4_4=y
+BR2_BINUTILS_VERSION_2_31_X=y
+BR2_GCC_VERSION_8_X=y
```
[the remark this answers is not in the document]
nickdesaulniers: Note that this change to this lone file has the commit message:

> * uClibc-ng shrinks the image sizes by a decent amount.

but it doesn't seem to change glibc related configs?

buildroot/arm64.config
```diff
@@ -1,7 +1,9 @@
 BR2_aarch64=y
 BR2_cortex_a57=y
 BR2_OPTIMIZE_3=y
-BR2_TOOLCHAIN_EXTERNAL=y
```
nickdesaulniers: any idea why this was set for arm64?

buildroot/arm.config
```diff
@@ -1,5 +1,8 @@
 BR2_arm=y
 BR2_OPTIMIZE_3=y
+BR2_KERNEL_HEADERS_4_4=y
+BR2_BINUTILS_VERSION_2_31_X=y
+BR2_GCC_VERSION_8_X=y
```
nickdesaulniers: github shows the binary size decreased by 99%, that can't be right, can it?

buildroot/arm64.config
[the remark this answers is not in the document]
nathanchance: > why upgrade binutils+gcc? Just to match ppc?

And x86_64 but yes, I'd like all configs to be identical where possible.

> Note that this change to this lone file has the commit message:
> 
> > * uClibc-ng shrinks the image sizes by a decent amount.
> 
> but it doesn't seem to change glibc related configs?

We were using an external toolchain with glibc in it so there was no config for it but that change is still happening. If I need to clean up the commit message, let me know.

buildroot/arm64.config
nathanchance: Was just the config I used when I first got into buildroot, I didn't want to build a toolchain in addition to the rootfs image.

buildroot/x86_64.config
nathanchance: Yes and arm64 implicitly (see above)

buildroot/arm.config
nathanchance: I think it means that 99% of it was rewritten but maybe not... I have zero idea.

nathanchance:
> > PowerPC should work with older kernels.
> 
> This comment brought to you pre-coffee.
> 
> Of course 4.4 doesn't have all of the clang fixes, so it won't build unless someone collects the patches and backports them.

By the way, I did backports for 4.14 and 4.19 so I can verify this works fine for PowerPC as well (I passed with an earlier patchset but I redid it with the commits from linux-next and I have to go to work but the builds should pass). I will look at 4.4 and 4.9 for all architectures in the next few days (if not, I'm off Monday and Tuesday again).

https://github.com/nathanchance/continuous-integration/commit/ccd30ea3e159e29185acd10ef5251ee4a246c8b1

https://github.com/nathanchance/continuous-integration/commit/816a5eac9656cbef29eb09cd6e5127a1a8e8dda2

https://travis-ci.com/nathanchance/continuous-integration/jobs/158894313

https://travis-ci.com/nathanchance/continuous-integration/jobs/158894309

nickdesaulniers:
ok, merged. Thanks for the clarifications.

PR #59, opened 2018-11-14
Enabling LLD for x86_64

tpimh:
Addressing the following issues:
- ~https://github.com/ClangBuiltLinux/linux/issues/29~
- ~https://github.com/ClangBuiltLinux/linux/issues/218~
- ~https://github.com/ClangBuiltLinux/linux/issues/31~
- ~https://github.com/ClangBuiltLinux/linux/issues/94~
- ~https://github.com/ClangBuiltLinux/linux/issues/30~

PR #60, opened 2018-11-15
Change LLVM repo to unstable to match Debian release

tpimh:


tpimh:
Hopefully with latest LLD #59 will stop failing, so it can be merged.

PR #65, opened 2018-11-19
patches: Add support for arm64 and x86_64 on 4.4

nathanchance:
This patchset is based on @mkaehlcke's: https://chromium.googlesource.com/chromiumos/third_party/kernel/+log/refs/sandbox/mka/llvm/v4.4

I have already verified this in my repo but it doesn't hurt to do it here too: https://travis-ci.com/nathanchance/continuous-integration/builds/91919320

I've pushed them to a git repo here to make it easier to review: https://github.com/nathanchance/linux/commits/linux-4.4.y

The final diff stat is not bad at all, hopefully Greg allows us to land this.

```
$ git diff --shortstat v4.4.163..
 23 files changed, 160 insertions(+), 136 deletions(-)
```

There is [one out of tree patch](https://github.com/nathanchance/linux/commit/7494b0450feaa89e9113c2ad05d200a49c8addf1), I think it is better than backporting [this series](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/arch/arm64?h=adf7589997927b1d84a5d003027b866bbef61ef2) that Mark Rutland did in 4.9 (plus any followup fixes and such).


.travis.yml
```diff
@@ -1,61 +1,10 @@
...
-    #- name: "ARCH=arm64 REPO=4.4"
-      #env: ARCH=arm64 REPO=4.4
-      #if: type = cron
-    #- name: "ARCH=x86_64 REPO=4.4"
-      #env: ARCH=x86_64 REPO=4.4
-      #if: type = cron
+    - name: "ARCH=arm64 REPO=4.4"
+      env: ARCH=arm64 REPO=4.4
+    - name: "ARCH=x86_64 REPO=4.4"
+      env: ARCH=x86_64 REPO=4.4
```
nickdesaulniers: assuming you're done testing as I'm marked for review (assign this PR to me when you're done to signal you're ready for review), we probably don't want to commit the deletions, and only want this for cron.

nathanchance: Thanks, done.

nathanchance:
Yeah it seems that I got a few conflicts early, like with patch 0007, which is absolutely needed for testing flags with `cc-option`. I'm happy to send out the patches via email tonight or I can wait until you hear back from Greg. Up to you.

PR #70, opened 2018-11-21
RISCV

martell:
ignore me just triggering travis in a PR


tpimh:
Currently to be fixed on LLVM side: https://github.com/ClangBuiltLinux/linux/issues/75 https://github.com/ClangBuiltLinux/linux/issues/76 (and many more not yet submitted)

tpimh:
Any progress with RISC-V?

martell:
It has been awhile, I should take a look at adding `medany` and `medlow` support to mcmodel flag in clang and then we can retry this, I will rebase for now given that the other issue has been resolved.

martell:
Need to 
- [ ] move what used to be the `debian-setup.sh` part into `https://github.com/ClangBuiltLinux/dockerimage/blob/master/Dockerfile`
Was thinking of waiting for #183 and #73 but it is not a big change and who knows how well / badly riscv is going to be supported by the llvm binutils replacements.
- [x] extract the buildroot part #187

nathanchance:
I will finish taking a look at this tonight when I am off work!

tpimh:
This is currently the only critical bug preventing RISC-V kernel from being compiled: https://github.com/ClangBuiltLinux/linux/issues/802

martell:
Looks like  https://github.com/ClangBuiltLinux/linux/issues/802 might not be an issue anymore.
I rebased on master but this is probably blocked on https://github.com/ClangBuiltLinux/dockerimage/pull/30

tpimh:
Great that you are back on it! In the meantime I have my own RISC-V branch here: https://github.com/ClangBuiltLinux/continuous-integration/compare/master...risc-v
I didn't yet add buildroot images or qemu boot testing. There are still some problems with RISC-V kernel.

nickdesaulniers:
@palmer-dabbelt had 3 patches needed to complete a build: https://git.kernel.org/pub/scm/linux/kernel/git/palmer/linux.git/log/?h=llvm

Then he reported the kernel fails to boot.
https://lore.kernel.org/linux-riscv/mhng-a55be814-523e-4aba-8b6d-f34d22eaee78@palmerdabbelt-glaptop1/

nickdesaulniers:
If anyone wants to try their hand at kernel debugging with GDB: https://nickdesaulniers.github.io/blog/2018/10/24/booting-a-custom-linux-kernel-in-qemu-and-debugging-it-with-gdb/

This would be a high value exercise for practice. ;)

nickdesaulniers:
Let's disable boot testing like we do with s390, and at least get build test coverage.

tpimh:
Good idea. Also I think we should at least print the output of `file vmlinux` (I believe both `riscv32` and `riscv64` targets now produce nearly identical 64 bit image). Maybe also add some more information about produced image if boot testing is disabled, possibly use `objdump`?

nathanchance:
@tpimh I think if we target just next for now, all of the patches you have in your branch are unnecessary. I’d say drop those, target next, then try a Travis run. If that all looks good, let’s ship it.

tpimh:
Seems fine: https://github.com/ClangBuiltLinux/continuous-integration/runs/512210549

buildroot/riscv32.config
```diff
@@ -0,0 +1,17 @@
...
+BR2_BINUTILS_VERSION_2_31_X=y
+BR2_GCC_VERSION_8_X=y
+BR2_TARGET_GENERIC_ROOT_PASSWD="root"
+BR2_TARGET_GENERIC_GETTY_PORT="ttyS0"
+BR2_ROOTFS_OVERLAY="overlays"
+BR2_PACKAGE_BUSYBOX_SHOW_OTHERS=y
+BR2_PACKAGE_BASH=y
+BR2_TARGET_ROOTFS_CPIO=y
+BR2_TARGET_ROOTFS_EXT2=y
+BR2_TARGET_ROOTFS_EXT2_4=y
```
nickdesaulniers: didn't we get rid of the ext4 images? Let's change to just make the CPIO images, and add them to the commit.

driver.sh
```diff
@@ -174,6 +174,28 @@ setup_variables() {
       image_name=bzImage
       qemu="qemu-system-x86_64" ;;
 
+    "riscv32")
+      config=defconfig
+      image_name=vmlinux
+      qemu="qemu-system-riscv32"
+      qemu_ram=512m
+      qemu_cmdline=( -machine virt
+                     -drive "file=images/riscv32/rootfs.ext4,format=raw"
```
nickdesaulniers: and them use them here

driver.sh
```diff
@@ -174,6 +174,28 @@ setup_variables() {
...
+      export ARCH=riscv
+      export CROSS_COMPILE=riscv64-linux-gnu- ;;
+
+    "riscv64")
+      config=defconfig
+      image_name=vmlinux
+      qemu="qemu-system-riscv64"
+      qemu_ram=512m
+      qemu_cmdline=( -machine virt
+                     -drive "file=images/riscv64/rootfs.ext4,format=raw"
```
nickdesaulniers: and here.

nathanchance:
Closing due to https://github.com/ClangBuiltLinux/continuous-integration/pull/238.

If we want riscv32 and/or QEMU booting, those can be done in separate pulls.

PR #71, opened 2018-11-23
If the kernel is linked with LLD, use it also as HOSTLD

tpimh:


driver.sh
```diff
@@ -103,7 +103,7 @@ mako_reactor() {
   KBUILD_BUILD_TIMESTAMP="Thu Jan  1 00:00:00 UTC 1970" \
   KBUILD_BUILD_USER=driver \
   KBUILD_BUILD_HOST=clangbuiltlinux \
-  make -j"${jobs:-$(nproc)}" CC="${CC}" HOSTCC="${CC}" LD="${LD}" "${@}"
+  make -j"${jobs:-$(nproc)}" CC="${CC}" HOSTCC="${CC}" LD="${LD}" HOSTLD="${LD}" "${@}"
```
nathanchance: How in the world did this work? If `LD` is not set, `LD` is defined based on `CROSS_COMPILE`... so we'd be host compiling but cross linking? Seems wrong to me even if it appears to work, I feel like HOSTLD should only be defined if `LD` == `lld`.

Something like this?

```diff
diff --git a/driver.sh b/driver.sh
index 3ca245e..dfdf258 100755
--- a/driver.sh
+++ b/driver.sh
@@ -102,11 +102,12 @@ mako_reactor() {
   KBUILD_BUILD_TIMESTAMP="Thu Jan  1 00:00:00 UTC 1970" \
   KBUILD_BUILD_USER=driver \
   KBUILD_BUILD_HOST=clangbuiltlinux \
-  make -j"${jobs:-$(nproc)}" CC="${CC}" HOSTCC="${CC}" LD="${LD}" "${@}"
+  make -j"${jobs:-$(nproc)}" CC="${CC}" HOSTCC="${CC}" LD="${LD}" HOSTLD="${HOSTLD:-}" "${@}"
 }
 
 build_linux() {
   CC="$(command -v ccache) $(command -v clang-8)"
+  [[ ${LD} =~ lld ]] && HOSTLD=${LD}
 
   if [[ -d ${tree} ]]; then
     cd ${tree}

```

tpimh: Thanks for your suggestion. Indeed it's very strange that the tests did succeed with linker targeting foreign architecture. I changed my commit to use your solution.

driver.sh
```diff
@@ -103,11 +103,12 @@ mako_reactor() {
   KBUILD_BUILD_TIMESTAMP="Thu Jan  1 00:00:00 UTC 1970" \
   KBUILD_BUILD_USER=driver \
   KBUILD_BUILD_HOST=clangbuiltlinux \
-  make -j"${jobs:-$(nproc)}" CC="${CC}" HOSTCC="${CC}" LD="${LD}" "${@}"
+  make -j"${jobs:-$(nproc)}" CC="${CC}" HOSTCC="${CC}" LD="${LD}" HOSTLD="${HOSTLD:ld}" "${@}"
```
nathanchance: `${HOSTLD:-ld}` should fix the Travis failure.

tpimh: Thank you!

tpimh:
Changed the commit message, should be good now.

tpimh:
What do you think about using LLD as HOSTLD always? It doesn't look like it can break things and if it can, we can switch back to bfd.

tpimh:
I never had a situation when linking with lld succeeded while linking with bfd failed, but it doesn't sound absolutely implausible either. Better just leave it like this: use lld as HOSTLD only if also using it for the target kernel.

PR #77, opened 2018-11-28
patches: Support PowerPC on 4.14 and 4.19

nathanchance:
These can't be pushed upstream just yet because the series that fixes the error in https://github.com/ClangBuiltLinux/linux/issues/259 isn't merged into mainline (it is sitting in linux-next currently) and the lib/raid6 change hasn't been accepted. Once those make their way into mainline, I will send these two series to stable so they can be dropped here but it doesn't hurt to get some early testing in :)

Presubmit results: https://travis-ci.com/nathanchance/continuous-integration/builds/92789308

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,9 @@ matrix:
     - name: "ARCH=arm64 REPO=4.19"
       env: ARCH=arm64 REPO=4.19
       if: type = cron
+    - name: "ARCH=ppc64le REPO=4.19"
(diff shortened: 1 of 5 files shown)


shenki:
Nice! Did you want to make the .travis.yml change in this PR, or do you plan to follow it up later?

nathanchance:
Whoops, that should be done now! Done, thanks for the review :)

PR #79, opened 2018-11-29
Next arm64

nickdesaulniers:

PR #82, opened 2018-12-01
Revert #79

nathanchance:
This has been fixed in a bit of a different way by Will Deacon upstream: https://git.kernel.org/arm64/c/efdb25efc7645b326cd5eb82be5feeabe167c24e

This commit has not appeared in -next just yet, I have only pushed this as a pull so that it is ready to merge when it does.


nathanchance:
The commit fixing this issue is now in -next: https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/?id=1b57ec8c75279b873639eb44a215479236f93481

nathanchance:
I didn't, I just happened to see so guess the notifications aren't working... Not really sure how we are supposed to test if it works (maybe another branch?). I'll try to do some research today and see what is wrong.

PR #83, opened 2018-12-02
patches: drop 4.19 arm patch that landed in 4.19.6

nickdesaulniers:


nickdesaulniers:
On the way into work this past week, I saw a license plate `+1 LGTM`.  Googlers have fun license plates.  Thanks for the review.

PR #87, opened 2018-12-02
arm64 Kaslr

nickdesaulniers:


ikitayama:
Is this Travis build a cross build?

nickdesaulniers:
@ikitayama yes; but that's because we're manually invoking the compiler with a target triple, example: `$ clang -target aarch64-linux-gnu foo.c`.  Nothing special in the Travis configs makes this a cross build.

The travis host machines are x86_64, so we aren't explicit about the target triple, defauting to the implicit target triple of the host.

ikitayama:
Nick, OK. do you happen to know if they soon deploy powerful, build farm-ready ThunderX2 or Neoverse N1?

ikitayama:
BTW, using LLD on ThunderX2 currently failing when taking up upstream kernel.

nathanchance:
> Nick, OK. do you happen to know if they soon deploy powerful, build farm-ready ThunderX2 or Neoverse N1?

We cannot comment on Travis's plans.

> BTW, using LLD on ThunderX2 currently failing when taking up upstream kernel.

I don't know that anyone has tried doing a native kernel build with LLVM on anything other than x86 (and maybe powerpc?). I'd be curious to see what the failure looks like, feel free to open an issue: https://github.com/ClangBuiltLinux/linux/issues/new

nickdesaulniers:
The LLD developers are (in my experience) exceeding fast to respond to bug reports.  If you could provide more information about how we can reproduce the issue, I'd be happy to put you in touch with them.

Can you please clarify, is it either that:
1. LLD is failing to run or link an upstream kernel when run on a thunderx2 host?
2. An upstream kernel that has been linked with LLD is not booting when run on a thunderx2?

We have some ThunderX2s at Google, I can like get access to one to reproduce if needed.

ikitayama:
1.  The build is failing on ThunderX2 at RIKEN.
Im motivated using it as LLD is supposed to be way faster than (go)ld.

nickdesaulniers:
In my experience, LLD is significantly faster than BFD and GOLD.

Can you please submit a bug report with more info in the link @nathanchance provided above? It would be helpful for us to know:

1. what kernel version (mainline, LTS)
2. any config in particular is related
3. the precise error message
4. what version of LLD you're using (as we may have already observed the error and fixed it in a newer version).

Thanks.

PR #89, opened 2018-12-03
ARM 4.9

nickdesaulniers:


patches/4.9/arm/clang-patches.4.9.arm.mbox
```diff
@@ -0,0 +1,492 @@
+From 642809370ba0aeae889137a05bbd250f445c1124 Mon Sep 17 00:00:00 2001
+From: Alistair Strachan <astrachan@google.com>
+Date: Mon, 3 Dec 2018 11:40:57 -0800
+Subject: [PATCH 1/8] [PATCH 1/5] efi/libstub: arm: support building with clang
```
nickdesaulniers: oops double patch lines

nickdesaulniers:
green run: https://travis-ci.com/ClangBuiltLinux/continuous-integration/builds/93425484

nickdesaulniers:
sent to stable: https://www.spinics.net/lists/stable/msg272526.html

nickdesaulniers:
I'm tempted to write a script that given a SHA, sees if it applies cleanly to LTS branches, and helps generates these mbox's but for multiple LTS branches. Need to think about it more.

@gregkh just merged these, so tonight's cron should fail, at which point, we can revert the patches/ change but keep the .travis build target.

nathanchance:
> I'm tempted to write a script that given a SHA, sees if it applies cleanly to LTS branches, and helps generates these mbox's but for multiple LTS branches. Need to think about it more.

That would be handy. I've just been cherry-picking on top of the latest tag then running `git format-patch --stdout v<version>.. > series.mbox`. I should see if I can autodetect `v<version>` so I could make a simple bash function like `stable-series`.

> @gregkh just merged these, so tonight's cron should fail, at which point, we can revert the patches/ change but keep the .travis build target.

Tonight should be fine since 4.9.143 won't be released for a day or two.

PR #90, opened 2018-12-04
PPC64 (big endian)

martell:
WIP

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -51,6 +51,17 @@ setup_variables() {
       qemu_cmdline=( -drive "file=images/x86_64/rootfs.ext4,format=raw,if=ide"
                      -append "console=ttyS0 root=/dev/sda" ) ;;
 
+    "ppc64")
+      config=pseries_defconfig
+      qemu="qemu-system-ppc64"
+      image_name=zImage.pseries
+      qemu_ram=1G
+      qemu_cmdline=( -machine pseries
+                     -nographic -vga none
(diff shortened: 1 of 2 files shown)


driver.sh
```diff
@@ -64,6 +64,19 @@ setup_variables() {
       export ARCH=powerpc
       export CROSS_COMPILE=powerpc64le-linux-gnu- ;;
 
+    "ppc64")
+      config=powernv_defconfig
```
shenki: Thanks for the patch!

The powernv defconfig is LE by default. We could build a BE powernv kernel, but instead we could get coverage of a different defconfig when booting BE. I suggest pseries.

We could build a pseries_defconfig, and boot it in a pseries qemu:
```
      config=pseries_defconfig
      qemu="qemu-system-ppc64"
      image_name=zImage.pseries
      qemu_ram=1G
      qemu_cmdline=( -machine s
                                  -nographic -vga none
                                  -initrd images/ppc64/rootfs.cpio )
```

shenki:
I did some further work on this https://github.com/shenki/continuous-integration/commits/powerpc64. There are some issues that I discovered, with 308 being a blocker:

- https://github.com/ClangBuiltLinux/linux/issues/35
- https://github.com/ClangBuiltLinux/linux/issues/240
- https://github.com/ClangBuiltLinux/linux/issues/308

tpimh:
With #196 merged this pull request is redundant.

PR #92, opened 2018-12-04
driver: arm64: change machine to allow testing KVM HYP

nickdesaulniers:
Speaking with arm64 reviewer Mark Rutland at Linux Plumbers conf 2018
about boot testing at EL2, this was the recommended change he had for us
to make sure we don't regress arm64 KVM HYP, which was an issue in the
past.

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -37,7 +37,7 @@ setup_variables() {
       image_name=Image.gz
       qemu="qemu-system-aarch64"
       qemu_ram=512m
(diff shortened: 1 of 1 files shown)


nathanchance:
Do we want to do boots at both EL1 and EL2?

nathanchance:
https://github.com/nathanchance/continuous-integration/commit/bd3d8c447fefa49b5e3f0bfe983e1b73c875edb2

PR #101, opened 2018-12-13
Add ARMv5 and ARMv6

shenki:
This renames the existing 32-bit ARM target to arm32_v7, and adds two more to test ARMv5 and ARMv6. A rootfs that can be shared by the other ISAs was added previously, so we don't need a new rootfs.


nathanchance:
> * [ ]  Work out how to build additional targets in order to have the device trees built. The solution in `driver: build default target` doesn't work, as building modules for the multi_v7_defconfig goes over the Travis time limit

Would the `dtbs` target not work?

> * [ ]  Look at refactoring the addtional config options (as is used to disable ftrace). We could instead allow targets to create their own fragments (or check fragments into the tree). For example:
> 
> ```
> echo "CONFIG_PPC_PSERIES=y" >> fragment.config
> echo "CONFIG_PPC_POWERNV=y" >> fragment.config
> echo "CONFIG_MTD=y" >> fragment.config
> echo "CONFIG_MTD_BLOCK=y" >> fragment.config
> echo "CONFIG_MTD_POWERNV_FLASH=y" >> fragment.config
> ARCH=powerpc scripts/kconfig/merge_config.sh \
>     arch/powerpc/configs/ppc64_defconfig \
>     arch/powerpc/configs/le.config \
>     fragment.config
> ```

Alternatively, have `CONFIGS_TO_APPEND` bash array that we iterate through that architectures can define.

groeck:
>     * [ ]  The guests cannot cause qemu to exit. I've submitted some qemu patches [[1](http://patchwork.ozlabs.org/patch/1010823/), [2](http://patchwork.ozlabs.org/patch/1010825/)]  to add support for this, I would need to backport them and do a new qemu build.
> 

I use "reboot" in the guest and the "-no-reboot" qemu command line option.

shenki:
> > ```
> > * [ ]  The guests cannot cause qemu to exit. I've submitted some qemu patches [[1](http://patchwork.ozlabs.org/patch/1010823/), [2](http://patchwork.ozlabs.org/patch/1010825/)]  to add support for this, I would need to backport them and do a new qemu build.
> > ```
> 
> I use "reboot" in the guest and the "-no-reboot" qemu command line option.

Thanks fort the suggestion. I've successfully switched to this for the ppc32 config, I'll do the same here.

driver.sh
```diff
@@ -19,17 +19,56 @@ setup_variables() {
...
   case ${ARCH:=arm64} in
-    "arm")
+    "arm32_v5")
+      config=multi_v5_defconfig
+      image_name=zImage
+      qemu="qemu-system-arm"
+      qemu_ram=512m
+      qemu_cmdline=( -machine palmetto-bmc
+                     -no-reboot
+                     -dtb "${REPO}/arch/arm/boot/dts/aspeed-bmc-opp-palmetto.dtb"
```
nathanchance: Changing `${REPO}` to `${tree}` here will be a little more future proof in case we start building these architectures on LTS branches: https://github.com/ClangBuiltLinux/continuous-integration/commit/7c4b607a5d36d155edff01df18e3f949df4b2c88

shenki: Okay. I added setting of tree in the switch statement, please take a look.

PR #104, opened 2018-12-20
Remove KASLR patches until this is sorted out on the LLD side.

nickdesaulniers:
Manual revert of 5e3e900b1c0a8cf6c96edab3ab331d3dd6e7f4fe


nickdesaulniers:
Looks like -next arm64 is not booting.  @nathanchance is there a known issue here?

nickdesaulniers:
stupid, I merged the do not merge commit. just pushed a revert straight to master.

nathanchance:
Looks like it's not a Clang issue: https://kerneltests.org/builders/qemu-arm64-next/builds/1035

Same thing happens with several arm builds, Guenter reported it already: https://lore.kernel.org/lkml/20181220183517.GA13028@roeck-us.net/

I'll keep an eye on it.

nathanchance:
I assume based on the stacktrace in Guenter's dashboard, this commit should fix it: https://git.kernel.org/pub/scm/linux/kernel/git/joro/iommu.git/commit/?id=dc9de8a2b20f495696330d60a289935f36407995

Let's see tonight.

nickdesaulniers:
Ah seems like a good idea to cross reference our CI failures with @groeck 's.

Though I think our failure is different.  If you look at the logs (grep+click `stdio`), there's clearly a panic, whereas our tests are failure to boot.  Unless that's what the first few lines of @groeck 's logs are showing?

groeck:
I am missing a bit of context, but AFAIK it is all the same problem. Some of my log output is silent because the crash happens before the console is activated. I have not been able to figure out how to enable earlycon with arm64 virt tests.

groeck:
Note though that arm64 (and pretty much everything else) has been flaky in -next for almost a month. There have been various issues over time. It almost looks like one gets fixed, two others are introduced. People seem to use -next as playground nowadays.

PR #107, opened 2019-01-01
Make everything green again

nathanchance:
Each commit message should be fairly self explanation. I chose not to break this down in to three different pull requests as none of the commits are particularly controversial and it will make everything pass in one go.


nathanchance:
Presubmit run: https://travis-ci.com/ClangBuiltLinux/continuous-integration/builds/96011365

PR #108, opened 2019-01-07
patches: arm32: Remove naked function usage patch

shenki:
This has been merged upstream as b99afae13901 ("ARM: 8805/2: remove
unneeded naked function usage").

Signed-off-by: Joel Stanley <joel@jms.id.au>


shenki:
This fixes the build failure for the arm32 targets.

PR #116, opened 2019-01-18
CI: 's/-8/-9/g'

nathanchance:
:) https://github.com/llvm/llvm-project/commit/1fe469ae6cd912248e4d754cc59c1231576a36eb

Depends on https://github.com/ClangBuiltLinux/dockerimage/pull/19

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -132,7 +132,7 @@ check_dependencies() {
   command -v ${qemu}
   command -v timeout
   command -v unbuffer
-  command -v clang-8
+  command -v clang-9
   command -v "${LD:="${CROSS_COMPILE:-}"ld}"
 }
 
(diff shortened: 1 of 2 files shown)


driver.sh
```diff
@@ -132,7 +132,7 @@ check_dependencies() {
   command -v ${qemu}
   command -v timeout
   command -v unbuffer
-  command -v clang-8
+  command -v clang-9
   command -v "${LD:="${CROSS_COMPILE:-}"ld}"
```
nickdesaulniers: is ld duplicated in this function?

would you mind adding a check for lld-9 while you're here?

nathanchance: >is ld duplicated in this function?

Oops, looks like it is... Do you want the duplicate one removed now or later?

>would you mind adding a check for lld-9 while you're here?

But `ld.lld` isn't strictly required to run this script. When we pass `LD=ld.lld-9` to the script on Travis, it will get verified with the `command -v "${LD:="${CROSS_COMPILE:-}"ld}"`

nickdesaulniers: good point!  Can be removed in this PR. ☃️

nathanchance: Done!

Issue #120, opened 2019-01-29
 bash[1] trap invalid opcode


shenki:
I used the container and driver.sh to build locally on my amd64 laptop, and then ran qemu manually with `init=/bin/bash` added to the command line. Unexpectedly it crashed:

```
qemu-system-x86_64 -m 512m -drive file=images/x86_64/rootfs.ext4,format=raw,if=ide \
 -append "console=ttyS0 root=/dev/sda init=/bin/bash" -nographic \
 -kernel linux/arch/x86_64/boot/bzImage
```
```
[   13.093363] Run /bin/bash as init process
[   13.257633] traps: bash[1] trap invalid opcode ip:47ed19 sp:7ffcf94e9680 error:0 in bash[400000+d0000]
[   13.266388] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000004
[   13.266761] CPU: 0 PID: 1 Comm: bash Not tainted 5.0.0-rc4+ #1
[   13.266905] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.12.0-1 04/01/2014
[   13.267230] Call Trace:
[   13.267510]  dump_stack+0xa7/0x10b
[   13.267672]  panic+0x104/0x2ed
[   13.267843]  ? do_exit+0x888/0x900
[   13.267998]  do_exit+0x900/0x900
[   13.268102]  do_group_exit+0xa0/0xc0
[   13.268218]  get_signal+0x638/0x8a0
[   13.268348]  ? rcu_read_lock_sched_held+0x86/0xc0
[   13.268490]  ? invalid_op+0xa/0x20
[   13.268603]  do_signal+0x2f/0x590
[   13.268710]  ? force_sig_info+0x153/0x160
[   13.268841]  ? invalid_op+0xa/0x20
[   13.268950]  prepare_exit_to_usermode+0xc7/0x160
[   13.269099]  retint_user+0x8/0x18
[   13.269318] RIP: 0033:0x47ed19
[   13.269506] Code: Bad RIP value.
[   13.269603] RSP: 002b:00007ffcf94e9680 EFLAGS: 00000202
[   13.269757] RAX: 00000000004a8277 RBX: 0000000000000001 RCX: 0000000000000000
[   13.269917] RDX: 00000000004e16e8 RSI: 0000000000000000 RDI: 00000000004e1b10
[...]

PR #121, opened 2019-01-29
buildroot: x86: Don't optimize for the corei7 variant

nathanchance:
Joel reported in #120 that when passing `init=/bin/bash` via the kernel
cmdline, there is a nice panic that happens on boot due to an invalid
opcode. Nick suggested that it was possible that the rootfs was compiled
with ISA extensions that aren't supported by QEMU's virtual CPU, which
is exactly the case.

When I initially created these images for myself, I selected
`CONFIG_BR2_x86_corei7=y` (as that's what I was using at the time),
which selects a bunch of ISA extensions [1]:

```
  config BR2_x86_corei7
      bool "corei7"
      select BR2_X86_CPU_HAS_MMX
      select BR2_X86_CPU_HAS_SSE
      select BR2_X86_CPU_HAS_SSE2
      select BR2_X86_CPU_HAS_SSE3
      select BR2_X86_CPU_HAS_SSSE3
      select BR2_X86_CPU_HAS_SSE4
      select BR2_X86_CPU_HAS_SSE42
```

I have never noticed a problem because I always use the `-cpu host
-enable-kvm` options, which makes QEMU's CPU emulate the host's (which
supported those extensions). QEMU's default cpu "qemu64" only supports
SSE3 [2], which Buildroot's default target variant "nocona" supports as
well. Use that variant and rebuild the images so there are no more
issues with `init=/bin/bash`.

Presubmit: https://travis-ci.com/nathanchance/continuous-integration/builds/99019841

[1]: https://github.com/buildroot/buildroot/blob/master/arch/Config.in.x86
[2]: https://github.com/qemu/qemu/blob/abd5fc4c862d/target/i386/cpu.c#L1488


nickdesaulniers:
Just a thought, but another solution is to always run qemu with the `-cpu host -enable-kvm` options.  Not sure if it slows down the build (doubt that it matters).  Thanks for figuring out that my hypothetical was indeed the case, and the report.

nathanchance:
I don't think `-cpu host -enable-kvm` works with Travis but I can try later. I think this is a fine solution for right now.

Thanks for the review!

PR #124, opened 2019-03-01
DO NOT COMMIT: try run

nickdesaulniers:


patches/linux/x86_64/0001-x86-boot-clean-up-headers.patch
```diff
@@ -0,0 +1,50 @@
...
+
+Because GCC performs inlining before semantic analysis, and
+__arch_hweight64 is dead in this translation unit, GCC does not report
+any issues at compile time.  Clang does the semantic analysis in the
+front end, before inlining (run in the middle) can determine the code is
+dead. I consider this another case of PR33587, which I think we can do
+more work to solve.
+
+It turns out that arch/x86/boot/string.c doesn't actually need
+linux/kernel.h, simply linux/limits.h and linux/types.h. Include them,
```
nickdesaulniers: should be `simply linux/limits.h and linux/compiler.h`.  `linux/types.h` was already included and was simply sorted, not added.

nathanchance:
This should pass when rebased on latest master. linux-next won't get any updates until March 15th so it'd be nice to get this merged so we can get all builds passing during that time (so that we can easily catch any regressions that happen during the merge window).

PR #125, opened 2019-03-03
patches: linux: x86_64: Update first patch for new conflicts

nathanchance:
Commit 2a418cf3f5f1 ("x86/uaccess: Don't leak the AC flag into
__put_user() value evaluation") upstream introduced some new conflicts.
That patch is a bug fix that shouldn't be reverted so update the revert
patches to work around it.

Unfortunately, mainline and linux-next have diverged enough for the
second patch to no longer apply cleanly to the other tree when based on
one, requiring a separate copy for each. We should be able to unify them
after 5.1-rc1.

Presubmit: https://travis-ci.com/nathanchance/continuous-integration/builds/102937246


nathanchance:
Note, #124 is required for next to pass with this, which I included in presubmit.

PR #127, opened 2019-03-06
enable llvm-ar by default

nickdesaulniers:


nathanchance:
Looks sane to me. There were no kernel patches needed right? I only ask because we're building back to 4.4.

nickdesaulniers:
I don't think so.  We could do a presubmit run; but I think it's ok to let it fail in postsubmit tonight and clean up tomorrow. WDYT?

nathanchance:
Sounds good to me. I am doing a run of linux-next builds on my box right now; once that is done, I'll do some tests locally and report back.

nickdesaulniers:
thanks for the review.  Will look into getting this turned on for Pixel kernels if everything is green (llvm-ar is already used in userspace).

PR #128, opened 2019-03-07
driver: Improvements around binary locating

nathanchance:
We've tailored driver.sh around our Docker image with regards to the
binary versions that we check for. A user might not have the latest
versions of Clang or llvm-ar installed. Check for all three versions
offered by apt.llvm.org, falling back to the generic binary name if
those are not found.

Additionally, for llvm-ar, do a sanity check that the user has a version
of llvm-ar that can actually build the kernel (has support for the P
modifier). If they don't fall back to regular GNU ar and let them know.

Presubmit: https://travis-ci.com/nathanchance/continuous-integration/builds/103470578

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -139,14 +139,72 @@ setup_variables() {
 }
 
 check_dependencies() {
+  # Check for existence of needed binaries
   command -v nproc
   command -v "${CROSS_COMPILE:-}"as
   command -v ${qemu}
   command -v timeout
   command -v unbuffer
-  command -v clang-9
-  command -v llvm-ar-9
-  command -v "${LD:="${CROSS_COMPILE:-}"ld}"
+
+  # Check for LD, CC, and AR environmental variables
+  # and print the version string of each. If CC and AR
+  # don't exist, try to find them.
+  # lld isn't ready for all architectures so it's just
+  # simpler to fall back to GNU ld when LD isn't specified
+  # to avoid architecture specific selection logic.
+
+  "${LD:="${CROSS_COMPILE:-}"ld}" --version
+
+  if [[ -z "${CC:-}" ]]; then
+    for CC in clang-9 clang-8 clang-7 clang; do
+      command -v ${CC} &>/dev/null && break
+    done
+  fi
+  ${CC} --version 2>/dev/null || {
+    set +x
+    echo
+    echo "Looks like ${CC} could not be found in PATH!"
+    echo
+    echo "Please install as recent a version of clang as you can from your distro or"
+    echo "properly specify the CC variable to point to the correct clang binary."
+    echo
+    echo "If you don't want to install clang, you can either download AOSP's prebuilt"
+    echo "clang [1] or build it from source [2] then add the bin folder to your PATH."
+    echo
+    echo "[1]: https://android.googlesource.com/platform/prebuilts/clang/host/linux-x86/"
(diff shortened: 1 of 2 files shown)


driver.sh
```diff
@@ -155,11 +193,12 @@ mako_reactor() {
...
   KBUILD_BUILD_USER=driver \
   KBUILD_BUILD_HOST=clangbuiltlinux \
-  make -j"${jobs:-$(nproc)}" CC="${CC}" HOSTCC="${CC}" LD="${LD}" HOSTLD="${HOSTLD:-ld}" AR="llvm-ar-9" "${@}"
+  make -j"${jobs:-$(nproc)}" CC="${CC}" HOSTCC="${CC}" LD="${LD}" HOSTLD="${HOSTLD:-ld}" AR="${AR}" "${@}"
 }
 
 build_linux() {
-  CC="$(command -v ccache) $(command -v clang-9)"
+  # Wrap CC in ccache
+  CC="$(command -v ccache) ${CC}"
```
nickdesaulniers: can we move this ccache check above?

driver.sh
```diff
@@ -139,14 +140,51 @@ setup_variables() {
...
+  if [[ -z "${AR:-}" ]]; then
+    for AR in llvm-ar-9 llvm-ar-8 llvm-ar-7 llvm-ar "${CROSS_COMPILE:-}"ar; do
+      ${AR} --version 2>/dev/null && break
+    done
+    # Optimistically check to see that the user has a llvm-ar
+    # with https://reviews.llvm.org/rL354044. If they don't,
+    # fall back to GNU ar and let them know.
+    if ${AR} --version | grep -q "LLVM" && \
+       [[ $(${AR} --version | grep version | sed -e 's/.*LLVM version //g' -e 's/[[:blank:]]*$//' -e 's/\.//g') -lt 900 ]]; then
+      set +x
```
nickdesaulniers: Can you put this check in another function with a descriptive name like `check_ar_version` or w/e?

driver.sh
```diff
@@ -139,14 +140,51 @@ setup_variables() {
...
+  # don't exist, try to find them.
+  # lld isn't ready for all architectures so it's just
+  # simpler to fall back to GNU ld when LD isn't specified
+  # to avoid architecture specific selection logic.
+
+  "${LD:="${CROSS_COMPILE:-}"ld}" --version
+
+  if [[ -z "${CC:-}" ]]; then
+    for CC in clang-9 clang-8 clang-7 clang; do
+      ${CC} --version 2>/dev/null && break
```
nickdesaulniers: what happens when no version is found?

driver.sh
nathanchance: `ccache` isn't a required binary, that's why it is here instead of up there. I'll make that a comment.

driver.sh
nathanchance: Yes.

driver.sh
nathanchance: Hmmm fair point, I'll need to add another check to make sure we fail here instead of during make.

nathanchance:
All comments should be addressed. If no `${CC}` or `${AR}` is found, the version command will now properly fail because we `set -e` at the top.

nathanchance:
I went ahead and added a more descriptive message if `clang` is not found (or if the user specified a non-existent binary in PATH).

PR #136, opened 2019-03-14
add facility to apply patches to all branches.

nickdesaulniers:
This probably wont get used too often, as patches probably wont apply
cleanly across all these different branches, but makes it simple to
apply patches to many trees.

Also, add patch for
https://github.com/ClangBuiltLinux/linux/issues/416

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -218,10 +218,11 @@ mako_reactor() {
 
 apply_patches() {
   patches_folder=$1
-  [[ -d ${patches_folder} ]] && git apply -v -3 "${patches_folder}"/*.patch
-  # Otherwise if the above conditional fails, a non-zero return code halts the
-  # build.
-  return 0
+  if [[ -d ${patches_folder} ]]; then
+    git apply -v -3 "${patches_folder}"/*.patch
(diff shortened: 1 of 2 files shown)


nathanchance:
I would also presubmit one architecture on each different repo/branch combo that we support, just to make sure the patch applied (I've tested arm64 on common 4.9 and stable 4.4).

nickdesaulniers:
odd, @nathanchance any idea why the build is stopping after applying the patches? Seems like the last output is `+ [[ -d ../patches/linux/arm ]]`.

nickdesaulniers:
https://stackoverflow.com/a/17336953/1027966:
```
The default return value is the exit value of the last statement executed within the function.
```
so I probably need an explicit `return 0` in there.

nathanchance:
I would do an `if else` with an explicit return zero in the else case, otherwise we won't capture the git apply error code.

nickdesaulniers:
green presubmit showing patch applies to each branch: https://travis-ci.com/ClangBuiltLinux/continuous-integration/builds/104603172

Issue #137, opened 2019-03-15
QEMU doesn't exit cleanly for PowerPC on Debian


nathanchance:
After switching us back to a Debian base in #135, the PowerPC builds do not exit cleanly. Reverting back to Ubuntu fixes this (which would be @shenki's package).

This doesn't appear to be Clang related as reverting back to `clang-8` doesn't change anything and I can't reproduce locally on my server, which is running Arch Linux and using QEMU 3.1 as well.

shenki:
Do you have a log?

nathanchance:
It just hangs until `timeout` kills it (exit code 124); sorry, should have been clearer.

shenki:
Any chance of a log so I can see where it gets to before hanging?

nathanchance:
Yup: https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/185056479

shenki:
Thanks. It's hanging when performing the shutdown:

```
The system is going down NOW!
Sent SIGTERM to all processes
Sent SIGKILL to all processes
Requesting system poweroff
[   21.196172] spin_lock-torture: Unscheduled system shutdown detected
[   21.206716] reboot: Power down
[   22.110546048,5] OPAL: Shutdown request type 0x0...
```
The shutdown request (IPMI message) doesn't take affect. This happens with a GCC built 4.19.13, so it's nothing to do with clang.

The version of skiboot, the powernv firmware, is v6.2. If I locally build skiboot 6.2 it fails. With skiboot master (v6.2-64-gccfd3547af37) it works. I did a scan of the 64 commits, and this one looked suspicious: https://github.com/open-power/skiboot/commit/44aa41061a9ffc289ff6b7d7b47132c4f15896bd. Sure enough, reverting that causes qemu to hang again.

We need to get that patch applied to the Debian skiboot build.

shenki:
I put together a fix where we include a newer skiboot along side the other ppc64le images:

 https://github.com/ClangBuiltLinux/continuous-integration/pull/138

I've also emailed the Debian maintainer to see if they can update the package. The latest skiboot-6.2.x release (6.2.2) includes the patch.

dileks:
Thanks @shenki

PR #138, opened 2019-03-15
ppc64le: Add skiboot image

shenki:
The Debian skiboot build fails to reboot or shutdown with Qemu 3.1. This
has been fixed in skiboot master but not yet in the Debian package.

The image added to the tree comes from the OpenPower CI system:

 https://openpower.xyz/job/openpower/job/openpower-op-build/label=slave,target=palmetto/

It is built from https://github.com/open-power/skiboot.

This change should be reverted when Debian releases a fixed package.

Resolves https://github.com/ClangBuiltLinux/continuous-integration/issues/137
Signed-off-by: Joel Stanley <joel@jms.id.au>


shenki:
Debian's bug tracker intimidates me. I sent some mail to the maintainer, so we'll see how that goes.

nickdesaulniers:
thanks @shenki !

PR #141, opened 2019-03-27
turn on objcopy by default for all targets

nickdesaulniers:


driver.sh
```diff
@@ -184,6 +183,14 @@ check_dependencies() {
   fi
   check_ar_version
   ${AR} --version
+
+  if [[ -z "${OBJCOPY:-}" ]]; then
+    for OBJCOPY in llvm-objcopy-9 llvm-objcopy-8 llvm-objcopy-7 llvm-objcopy "${CROSS_COMPILE:-}"objcopy; do
+      command -v ${OBJCOPY} 2>/dev/null && break
+    done
+  fi
+  check_ar_version
```
nathanchance: Should be `check_objcopy_version` :)

nickdesaulniers:
rerunning, as I think https://github.com/ClangBuiltLinux/linux/issues/435 is now resolved.

nathanchance:
I don't think we don't have that patch in our clang build yet (version string says r360126, the patch is r360233). It should be there on the next refresh.

nickdesaulniers:
@nathanchance I think we can merge this. Pre-existing failures are related to lld.

nathanchance:
Hmmm interesting, how did this pass? Clang wasn't refreshed last night according to apt.llvm.org so nothing should have changed...

nickdesaulniers:
oh, we probably fail the `-n` lld check for arm64 before we hit the objcopy issue.

Issue #27, opened 2018-11-06
more configs


tpimh:
(not a duplicate of #21)

I think additional configs may be tested on travis. `defconfig` is good, but it also makes sense testing `tinyconfig` (+ enable elf support and tty driver, it will run init and print boot log to console), `allyesconfig` and `allmodconfig`. And run all in parallel with travis. Will it be good for testing or just a waste of resources?

arndb:
allmodconfig and allyesconfig are generally intended for build testing, not runtime testing. Getting allyesconfig to boot would be a lot of work, though I have managed to boot an x86 allmodconfig (after disabling CONFIG_MAXSMP and CONFIG_CMDLINE_OVERRIDE), and Anders Roxell has managed to boot arm64 allmodconfig after overriding around a dozen different options. This was all using gcc, so I'd expect additional problems with clang.
Getting those to work in clang would be great of course, but I'm not sure about running them continuously unless we first make it reliable on gcc.

nathanchance:
Not to mention that Travis is our biggest bottleneck. We can't even do a full defconfig build in the allotted time (50 minutes), we're just building the kernel image rather than all of the modules as well.

nickdesaulniers:
I can ask my boss if we can pay for beefier travis machines, but I think kernel-ci may be better for heavier lifting.

I wonder if we can have more than one cron jobs?  For example, I wonder if we could do daily builds and per pull request builds of a subset of targets, then a weekly build of all targets?

nathanchance:
Yeah I am sure Travis would let us break down the matrix into different subsets that we can target with different actions.

tpimh:
I contacted Travis support and they said that they can increase time limit to 180 minutes by request, but no more than this.

nickdesaulniers:
@tpimh neat, good to know.  Thanks for doing that.  I wonder if they have a limit on the size of our testing "matrix."  Because I see that as being the limiting factor for the number of configs we can test in a given run.

tapaswenipathak:
Reading the repository I think there is only common and tt config right now. What is this ticket's status on `allyesconfig` and `allmodeconfig`? This seems interesting, let me know if any of you is already working on it.

nickdesaulniers:
I expect an allyesconfig to timeout on travisCI (thus making that config untestable).  TravisCI gives us 50 minutes per test before it considers a build failed/timed out.

nathanchance:
[...]

PR #143, opened 2019-03-31
continuous-integration: Add allyesconfig

tapaswenipathak:
I mailed Travis about increasing the limit.

tl;dr this should close https://github.com/ClangBuiltLinux/continuous-integration/issues/27.

- [ ] @nathanchance comment
- [ ] Updates from Travis


nathanchance:
`allyesconfig` is a build target, it should be called by `mako_reactor allyesconfig` here: https://github.com/ClangBuiltLinux/continuous-integration/blob/ded712131712186b766273eea1761ff18495217e/driver.sh#L250

arm64 (and maybe x86_64) on linux-next will be the only targets that build successfully with it though.

tpimh:
Any progress on this one? There has been changes to the build system, this should be rebased.

nathanchance:
I still do not believe this will be a good addition to our matrix. Our latest build took almost 9 hours to complete because more and more files get built with each kernel release. Adding allyesconfig to this will make it take significantly longer because we can only have three builds going at a time. For that matter, we only get 180 minutes maximum per job and an x86 allyesconfig build on a four core machine ([Packet's t1.small.x86](https://www.packet.com/cloud/servers/t1-small/)) with LLVM 10 on Ubuntu 18.04 takes around 5 hours and requires at least 8GB of disk space for the kernel source and resulting files (we only get two cores and 18GB of disk space). I don't think Travis will be able to handle this.

tpimh:
Maybe changes should be made only to driver.sh and not affect our travis matrix. I would suggest introducing `CONFIG` variable which can be specified like `REPO`. If unspecified, the same old set of defconfigs is used, if set to "allyesconfig", builds allyesconfig. So it would be extremely easy to test locally by just running `./driver.sh CONFIG=allyesconfig` and waiting for a while.

nathanchance:
Yes that would be okay. We could also come up with a list of common configs that tend to stress issues in LLVM/Clang and build with those (some immediate ones are CONFIG_RANDOMIZE_BASE and CONFIG_KASAN).

PR #146, opened 2019-04-07
patches: linux: all: Drop bcmp patch

nathanchance:
It has been merged into mainline: https://git.kernel.org/linus/5f074f3e192f10c9fade898b9b3b8812e3d83342

Fixes the following Travis failure: https://travis-ci.com/ClangBuiltLinux/continuous-integration/builds/107314169


nickdesaulniers:
should we wait for this to hit stable?

nathanchance:
It might take a week or so for this to get queued up then released and I don't think we want CI to be held up for that long.

nathanchance:
Thanks :)

PR #148, opened 2019-04-11
driver: Dump the comment section of vmlinux

nathanchance:
This will help verify that LLD is working for linking.

```
+ readelf --string-dump=.comment vmlinux

String dump of section '.comment':
  [     0]  clang version 9.0.0 (git://github.com/llvm/llvm-project c84107612ad7c73679996b6b0fb39a7423f0058d)
  [    62]  Linker: LLD 9.0.0 (git://github.com/llvm/llvm-project c84107612ad7c73679996b6b0fb39a7423f0058d)
```

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -141,6 +141,10 @@ check_dependencies() {
   command -v timeout
   command -v unbuffer
 
+  for readelf in llvm-readelf-9 llvm-readelf-8 llvm-readelf-7 llvm-readelf; do
+    command -v ${readelf} &>/dev/null && break
+  done
+
   # Check for LD, CC, and AR environmental variables
   # and print the version string of each. If CC and AR
   # don't exist, try to find them.
@@ -259,6 +263,7 @@ build_linux() {
   mako_reactor olddefconfig &>/dev/null
   mako_reactor ${image_name}
   [[ $ARCH =~ arm ]] && mako_reactor dtbs
+  ${readelf} --string-dump=.comment vmlinux
 
   cd "${OLDPWD}"
 }


driver.sh
```diff
@@ -142,6 +142,7 @@ check_dependencies() {
   # Check for existence of needed binaries
   command -v nproc
   command -v "${CROSS_COMPILE:-}"as
+  command -v readelf
```
nickdesaulniers: why you no use `llvm-readelf`??? :see_no_evil: :monkey: :monkey_face:

nathanchance: TIL that was a thing :P

We would need check for the different available versions like `clang` and `llvm-ar` though but that'd not hard.

nathanchance:
Updated with `llvm-readelf` first.

driver.sh
```diff
@@ -141,6 +141,11 @@ check_dependencies() {
   command -v timeout
   command -v unbuffer
 
+  for readelf in llvm-readelf-9 llvm-readelf-8 llvm-readelf-7 readelf; do
```
nickdesaulniers: I think this should have another `llvm-readelf` at the end (un-versioned, no GNU readelf)?

nathanchance: Thanks for catching that, I don't know how that happens because I tested in both our Docker container and my server's regular environment :/

driver.sh
```diff
@@ -141,6 +141,11 @@ check_dependencies() {
   command -v timeout
   command -v unbuffer
 
+  for readelf in llvm-readelf-9 llvm-readelf-8 llvm-readelf-7 llvm-readelf readelf; do
```
nickdesaulniers: thanks for the addition of `llvm-readelf`.  Can you additionally remove `readelf`?  I'd like to get to the point where we can build in an environment that contains no GNU binutils.  If `llvm-readelf` fails for whatever reason, than that's a bug we should fix.

nathanchance: `driver.sh` should be environment agnostic. I'm fine with making that change but if `llvm-readelf` isn't installed, we need to avoid trying to call it so I'll need to rework that.

nathanchance: Maybe it is worth adding an environment variable that could be exported/set to disable GNU fallbacks like `NO_GNU`.

nickdesaulniers: The ultimate goal needs to be that we can build with no GNU tools installed.  That way if any of the kernel build steps are not hermetic (ex. they don't invoke they linker they were told to) they fail hard.  That's a feature, not a bug.

nathanchance: Yes, I agree. If that is the direction you want to take it now, that's fine with me but that means that `driver.sh` will be exclusively tied to our CI setup because we can't expect people to remove their copy of binutils to use it.

Otherwise, the alternative is to make `env-setup.sh` either remove or check for the absence of GNU packages (or avoiding installing them during the docker build).

nickdesaulniers: Those are all good ideas and good points.  We can cross that bridge when we get there (explicitly checking that the env is does not have GNU binutils).  I think for this, expecting llvm-readelf to be installed (and lld and llvm-ar and llvm-objcopy etc) is not too demanding.

nathanchance: Indeed. Most distributions will include the LLVM tools when installing clang (or make them very easy to get otherwise).

I have removed `readelf` from the list so that the build fails when `llvm-readelf` isn't found.

nathanchance:
Thanks for the review and insight as always!

PR #149, opened 2019-04-11
driver: Use bash's default assignment operator for qemu_ram

nathanchance:
qemu_ram only needs to be assigned to something different if the default
causes issues. Move the default assignment to its first use, which will
make adding new architectures this script easier as there is one less
variable to worry about.


nathanchance:
Thanks for the review!

PR #163, opened 2019-05-13
enable clang's integrated assembler for arm64

nickdesaulniers:


nathanchance:
Will this actually work with `-no-integrated-as` being added to both `KBUILD_CFLAGS` and `KBUILD_AFLAGS`?

nickdesaulniers:
oh! good point! (I should verify locally w/o aarch64-linux-gnu-as)

nickdesaulniers:
https://github.com/ClangBuiltLinux/linux/issues/473

nathanchance:
For the record, this now works as expected after the patch that resolves https://github.com/ClangBuiltLinux/linux/issues/577.

Still will fair because of https://github.com/ClangBuiltLinux/linux/issues/473.

driver.sh
```diff
@@ -191,6 +191,12 @@ check_dependencies() {
   fi
   check_objcopy_version
   ${OBJCOPY} --version
+
+  if [[ -z "${AS:-}" ]]; then
+    for AS in clang-9 "${CROSS_COMPILE:-}"as; do
```
tpimh: This always prefers clang as `AS` unless `AS` environment variable is specified.
There are two ways to change it:
- Specify `AS` in similar fashion as `LD` is specified (preferred)
- Basically invert `AS` variable in travis environment: for every job which specifies `AS` remove the variable, for every job without this variable add is with value set to appropriate `as` name for cross-compilation

Anyway this wouldn't work at all because the variable isn't passed to `make`. I fixed it here: https://github.com/ClangBuiltLinux/continuous-integration/commit/1a2921f333043271eb3c9ec3808fd91ebe4ca6a0#diff-edd6bdbc5716dde5e06b45ae44b43caaR313

PR #166, opened 2019-05-24
driver: Disable LTO/CFI by default

nathanchance:
This is now enabled by default on Cuttlefish 4.19, which causes issues
with our Debian setup because we do not have llvm-nm installed and it is
hardcoded that way in the Makefile:

https://github.com/aosp-mirror/kernel_common/blob/50f91435a27a9f7/Makefile#L617

It might be worth making that customizable in the future; however,
Travis cannot handle LTO for both time and memory reasons so just
disable it by default. A user can still pass '--lto' to driver.sh
to avoid disabling LTO (it will NOT attempt to enable it).

Fixes: https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/202781149

Presubmit: https://travis-ci.com/nathanchance/continuous-integration/builds/113113278

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -9,6 +9,7 @@ setup_variables() {
       "-c"|"--clean") cleanup=true ;;
       "-j"|"--jobs") shift; jobs=$1 ;;
       "-j"*) jobs=${1/-j} ;;
+      "--lto") disable_lto=false ;;
       "-h"|"--help")
         cat usage.txt
         exit 0 ;;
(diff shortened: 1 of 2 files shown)


nickdesaulniers:
> because we do not have llvm-nm installed

eh? why not?

nathanchance:
Sorry, should have been clearer. `llvm-nm` is not installed, `llvm-nm-9` is.

nathanchance:
I've updated the commit message to clarify this and offer up a solution on the kernel/common side but this has to be done regardless of that. Merging now.

Issue #167, opened 2019-05-31
LLVM_VERSION=8 builds after asm goto


nathanchance:
This is a meta issue to decide what to do about the LLVM 8 builds we are running once asm goto is available to us in an LLVM 9 build. Do we keep the asm goto removal patches around for just LLVM 8 or do we turn off LLVM 8 builds until LLVM 9 is stable, where we would make LLVM 9 builds the stable ones and LLVM 10 the development ones?

nickdesaulniers:
implications for https://github.com/ClangBuiltLinux/continuous-integration/commits/asm_goto.  Let me think more about this.  @arndb has a crazy idea to do our own release of clang-8 for kernel.org, with asm goto fully backported.  I like that idea.

nathanchance:
![he's a madman](https://www.reactiongifs.us/wp-content/uploads/2013/08/hes_a_madman_futurama.gif "He's a madman!")

I like the idea too, especially since it would tackle getting clang builds available for everyone which was mentioned back at Linux Plumbers.

PR #170, opened 2019-06-04
patches: 4.19: x86_64: Update after 4.19.48 release

nathanchance:
The compiler.h patch has been merged [1].

To avoid arch/x86/boot/ warnings with clang, I backported Masahiro's
JUMP_LABEL and CC_HAS_ASM_GOTO refactoring [2], which causes conflicts
in our out of tree patch. We can now use the same patch as mainline
(which will include the improvement from #119), which we will hopefully
be able to drop soon!

[1]: https://git.kernel.org/stable/c/753328727cab4116410f2a3c150826ab6c555d6a
[2]: https://git.kernel.org/stable/c/0276ebf16675f3745d8b6f384dcf917e1379eda2


nickdesaulniers:
sorry, now that asm goto has landed, can't/shouldn't we just drop all of patches via merging #171

nathanchance:
Yes, we should; however, apt.llvm.org's clang hasn't been updated for over a week at this point (almost two). I haven't made a big deal of it because asm goto was still in the works but now that it has landed (and hopefully sticks), we should remove all of  our hacks.

I think that this can be merged now then once clang has been updated, we merged #171. Additionally, we need to think about #167 when merging #171.

nathanchance:
Thanks for the review!

PR #172, opened 2019-06-04
build-all.sh: Initial revision

nathanchance:
This is a script that will run all of the builds that Travis does on a
local user's machine. This can help validate changes faster than Travis
can if the user has a very fast machine (like I do).

### DIFF
--- a/build-all.sh
+++ b/build-all.sh
@@ -15,6 +15,9 @@ function format_time() {(
 )}
 
 BLD_ALL_START=$(date +%s)
+GREEN="\033[01;32m"
+RED="\033[01;31m"
+RESET="\033[0m"
 
 echo
 grep "env:" .travis.yml | grep -v LLVM_VERSION | sed "s/.*env: //g" | while read -r ITEM; do (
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
> if the user has a very fast machine (like I do).

weird flex but :ok_hand:

build-all.sh
```diff
@@ -0,0 +1,42 @@
...
+    [[ -n ${HOURS} ]] && TIME_STRING="${HOURS}h "
+    echo "${TIME_STRING}${MINS}m ${SECS}s"
+)}
+
+BLD_ALL_START=$(date +%s)
+
+echo
+grep "env:" .travis.yml | grep -v LLVM_VERSION | sed "s/.*env: //g" | while read -r ITEM; do (
+    DRIVER_START=$(date +%s)
+    read -ra VALUES <<< "${ITEM}"
```
nickdesaulniers: `<<<` will fail on dash shell

I learned this the hard way recently when Android kernel builds were failing, but only due to the build machines using `dash` while the internal "glinux" distros used `bash` (and I use `zsh`).

build-all.sh
```diff
@@ -0,0 +1,42 @@
...
+        export "${VALUE:?}"
+    done
+
+    # Make sure that the version suffix is stripped for local builds,
+    # where it is assumed that the user will be using a tip of tree build
+    [[ -n ${LD} ]] && export LD=${LD//-*}
+
+    echo -e "Running '${ARCH:+ARCH=${ARCH} }${LD:+LD=${LD} }${REPO:+REPO=${REPO} }./driver.sh'... \c"
+    if ! ./driver.sh "${@}" &>/dev/null; then
+        echo -e "\033[01;31mFailed\033[0m\c"
```
nickdesaulniers: please use variables for what I assume is green/red/reset

build-all.sh
nathanchance: Hmmm, the shebang specifies `bash` so is this a problem? I'll try to look at a more portable solution though.

build-all.sh
nathanchance: Sure.

build-all.sh
nathanchance: Well, `dash` doesn't do arrays so this entire script will have to be rewritten if being portable is a requirement. `driver.sh` uses some `bash`-isms too, I don't think it is unreasonable to require `bash` for these scripts. Let me know if you feel strongly about that though :)

PR #178, opened 2019-06-10
patches: llvm-8: Remove stray patch file

nathanchance:
Don't know how I didn't catch this :( doesn't affect anything because it
will never be called but annoying to have this in the wrong place
regardless.


nathanchance:
Sounds good, thanks!

PR #190, opened 2019-07-10
buildroot: Make it easier to rebuild all of the images after an upgrade/change

nathanchance:
Commit messages should be fairly self explanatory, let me know if there are any questions/concerns!


MSF-Jarvis:
I was trying to run this for the buildroot upgrade PR earlier and never quite figured out how, this is a welcome change :hugs:. Trying it out right now, will submit review once done.

MSF-Jarvis:
@nathanchance I can't figure out why my build fails, anything obvious I may have missed? https://del.dog/raw/qiqopafori

nickdesaulniers:
> make[1]: Leaving directory '/home/msfjarvis/git-repos/continuous-integration/buildroot/src/output/build/host-gcc-initial-8.3.0/build'
> output/images/rootfs.cpio could not be found! Did the build error?

MSF-Jarvis:
> make[1]: Leaving directory '/home/msfjarvis/git-repos/continuous-integration/buildroot/src/output/build/host-gcc-initial-8.3.0/build'
> output/images/rootfs.cpio could not be found! Did the build error?

Yeah Nathan helped me find the problem over Telegram, for some reason my C++ compiler was there but configure refused to use it.

nathanchance:
Thanks! @nickdesaulniers did you have any comments?

buildroot/rebuild.sh
```diff
@@ -1,44 +1,62 @@
...
+        make menuconfig
+        make savedefconfig
     fi
-    cp -v "${IMAGE}" "${IMAGES_FOLDER}"
+
+    # Build images
+    make -j"$(nproc)"
+
+    # Make sure images folder exists
+    IMAGES_FOLDER=../../images/$(basename "${CONFIG//.config}")
```
nickdesaulniers: was the double `//` intentional?

nathanchance: Yes, it's a habit (`/` removes the first occurrence, `//` removes all occurrences). Here it obviously does not matter so I can change it if it helps make it clearer.

buildroot/rebuild.sh
```diff
@@ -5,13 +5,21 @@
...
+    cd src || exit 1
+    if [[ $(git describe --exact-match --tags HEAD) != "${BUILDROOT_VERSION}" ]]; then
+        git fetch origin ${BUILDROOT_VERSION}
+        git checkout ${BUILDROOT_VERSION}
+    fi
+
+    # Clean up artifacts from the last build
+    make clean
+else
+    git clone -b ${BUILDROOT_VERSION} git://git.busybox.net/buildroot src
```
nickdesaulniers: should be a shallow clone?

nickdesaulniers:
> This will save bandwidth across image upgrades/rebuilds.

How's that?  Seems like a compressed tarball would be smaller than all of git history?

buildroot/rebuild.sh
nathanchance: Sure, I can make that change.

nathanchance:
> > This will save bandwidth across image upgrades/rebuilds.
> 
> How's that? Seems like a compressed tarball would be smaller than all of git history?

I guess I was comparing it to redownloading the tarball over and over in the current scheme. I suppose what I could do is just not remove the downloaded tarball source every rebuild and use the `print-version` make target to check if we need to update the source. At the same time, shallow clone + new tags is a little bit easier yet should achieve the same thing.

buildroot/rebuild.sh
nickdesaulniers: sorry, what do you mean by "removes" ?

buildroot/rebuild.sh
nickdesaulniers: That may change the fetch case to:
```
$ git fetch && git reset --hard FETCH_HEAD
```

nickdesaulniers:
> I guess I was comparing it to redownloading the tarball over and over in the current scheme.

Ah, I see.  In that case, I think it's simpler to wrap the fetch in a check if the file does not already exist, then fetch it.  Might be nice to check the checksum of it either way if buildroot provided them.

buildroot/rebuild.sh
nathanchance: Sorry, removes the part after the slashes from the variable's value (an example makes it clearer):

```
$ CONFIG=arm.config.config; echo ${CONFIG/.config}
arm.config
$ CONFIG=arm.config.config; echo ${CONFIG//.config}
arm
```

nickdesaulniers: oh, maybe create a new variable on the line before use with a comment; I feel like I'll forget that's what `//` does in the future. Thanks for the explanation.

nathanchance:
Alright, I think I have addressed all comments :)

buildroot/rebuild.sh
```diff
@@ -21,15 +26,14 @@ done
 # Download latest buildroot release
 BUILDROOT_VERSION=2019.02.3
 if [[ -d src ]]; then
-    cd src || exit 1
-    if [[ $(git describe --exact-match --tags HEAD) != "${BUILDROOT_VERSION}" ]]; then
-        git fetch origin ${BUILDROOT_VERSION}
-        git checkout ${BUILDROOT_VERSION}
+    if [[ $(cd src && make print-version | cut -d - -f 1 >/dev/null) != "${BUILDROOT_VERSION}" ]]; then
```
nickdesaulniers: if you redirect stdout from `cut` to `/dev/null`, is the subshell ever going to return a string to compare against `"${BUILDROOT_VERSION}"`?

buildroot/rebuild.sh
```diff
@@ -3,7 +3,16 @@
 
 function download_br() {
     mkdir -p src
-    curl https://buildroot.org/downloads/buildroot-"${BUILDROOT_VERSION}".tar.gz | tar -xzf - -C src --strip-components=1
+    TARBALL=buildroot-${BUILDROOT_VERSION}.tar.gz
+    rm -rf "${TARBALL}"
```
nickdesaulniers: probably don't need `-r` for a tarball

buildroot/rebuild.sh
```diff
@@ -3,7 +3,16 @@
...
+    curl -LO https://buildroot.org/downloads/"${TARBALL}"
+    SHA256=31a0dd7f5df75d20f6b74f0684918b643e5bf08d87dd5f9f02525852eecccf95
```
nickdesaulniers: add it as a file, then use `sha256sum -c <file>`
see: https://shapeshed.com/unix-sha1sum/

buildroot/rebuild.sh
nathanchance: Ah whoops, this should be `2>/dev/null` :(

buildroot/rebuild.sh
nathanchance: Yes, true.

buildroot/rebuild.sh
nathanchance: Sure

nathanchance:
Thanks for the review all!

PR #192, opened 2019-07-27
Add support for MIPS 32-bit little endian

nathanchance:
Tested using a build of LLVM at https://github.com/llvm/llvm-project/commit/b45631090220b732e614b5530bbd1d230eb9d38e and https://github.com/nathanchance/continuous-integration/commit/0cdb1641521cb1413e25f11220c35e22dcdcc5be

Build and boot log: https://gist.github.com/b5f8e6d0d8f0a5ad11379fbff8894071


driver.sh
```diff
@@ -292,7 +304,14 @@ build_linux() {
 }
 
 boot_qemu() {
-  local kernel_image=${tree}/arch/${ARCH}/boot/${image_name}
+  local kernel_image
+
+  if [[ ${image_name} = "vmlinux" ]]; then
```
nickdesaulniers: What's up with `image_name`? Does the mips build not place the resulting image in `arch/mips/boot/vmlinux`?

nathanchance: No. `vmlinux` is just the raw kernel image, which always ends up in the root of the kernel or object directory. `malta_defconfig` appears to generate a binary kernel image `vmlinux.bin` in `arch/mips/boot/compressed` but it seems like most people just use the regular kernel image.

nickdesaulniers: Are there no kernel binary images in `arch/mips/boot/` (not `compressed`)?

nathanchance:
```
$ fd --no-ignore --maxdepth 1 . out/arch/mips/boot
out/arch/mips/boot/compressed
out/arch/mips/boot/dts
out/arch/mips/boot/tools
```

nickdesaulniers: what if you set:

```
image_name = "compressed/vmlinux.bin.z"
```
in the mips block, then you don't need this???

nathanchance: I'll see if that works.

driver.sh
```diff
@@ -88,21 +88,17 @@ setup_variables() {
...
-      [[ $(grep -c -E 'vmx|svm' /proc/cpuinfo) -gt 0 && $(lsmod 2>/dev/null | grep -c kvm) -gt 0 ]] && qemu_cmdline=( "${qemu_cmdline[@]}" -enable-kvm )
-      image_name=bzImage
-      qemu="qemu-system-x86_64" ;;
+    "mipsel")
+      config=malta_defconfig
+      image_name=vmlinux
+      qemu="qemu-system-mipsel"
+      qemu_cmdline=( -machine malta
+                     -cpu 24Kf
+                     -append "root=/dev/sda"
```
nickdesaulniers: probably need to append a console too? oh, looks like no...hmm...interesting.

driver.sh
nathanchance:
```
+ timeout 2m unbuffer qemu-system-mipsel -m 512m -machine malta -cpu 24Kf -append root=/dev/sda -drive file=images/mipsel/rootfs.ext4,format=raw,if=ide -display none -serial mon:stdio -kernel linux/arch/mips/boot/vmlinux.bin
qemu-system-mipsel: could not load kernel 'linux/arch/mips/boot/vmlinux.bin': The image is not ELF
```

nickdesaulniers: yep...strange inconsistency, but w/e

PR #193, opened 2019-07-27
CI: Update LLVM version

nathanchance:
To mirror https://github.com/llvm/llvm-project/commit/8f5b44aead89a56c6fbf85ccfda03ae1e82ac431.

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -144,7 +144,7 @@ check_dependencies() {
   command -v timeout
   command -v unbuffer
 
-  for readelf in llvm-readelf-9 llvm-readelf-8 llvm-readelf-7 llvm-readelf; do
+  for readelf in llvm-readelf-10 llvm-readelf-9 llvm-readelf-8 llvm-readelf-7 llvm-readelf; do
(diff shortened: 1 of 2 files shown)


nathanchance:
I have an idea that might work.

driver.sh
```diff
@@ -144,7 +150,10 @@ check_dependencies() {
   command -v timeout
   command -v unbuffer
 
-  for readelf in llvm-readelf-9 llvm-readelf-8 llvm-readelf-7 llvm-readelf; do
+  oldest_llvm_version=7
+  latest_llvm_version=10
```
nickdesaulniers: so my point is that a human has to write up a patch whenever a new version of llvm comes out.

I'm [lazy](http://wiki.c2.com/?LazinessImpatienceHubris) and don't want to have to do that; we should come up with a solution where upon no human intervention, this code continues to just work for new llvm versions, without having to have a human come and manually patch this script.

nathanchance: I'll have to brainstorm more on how to do this...

However, I think this is better than before as we will only need to update this one location.

PR #194, opened 2019-07-28
Unbreak ppc32

nathanchance:
Message for first commit:

Currently, we collapse arm32_v{5,6,7} and ppc32/powerpc64le into arm and
powerpc respectively. We don't want to apply patches that are for only
one specific sub-architecture to all so add a SUBARCH variable that can
be used to differentiate everything.

Message for second commit:

It does not seem like this is going to be dealt with in a timely manner;
rather than disabling the target outright, just revert the one spot that
breaks everything.

Presubmit: https://travis-ci.com/nathanchance/continuous-integration/builds/121137965


nathanchance:
It's probably worth making this a clang only thing like you did, I'll update that + the commit message.

nickdesaulniers:
> It's probably worth making this a clang only

Sorry, yes I just thought of that, too.  Please run it through the CI to make sure just `dcbz` is the only problematic function.

nathanchance:
Hmmm, thinking about this a little bit more, I think I like your patch in #195 a little bit more than mine because all of these functions have this code generation issue:

https://godbolt.org/z/z57VU7

While `dcbz` might be the only problematic one at the moment, one of the other ones might be at another point in time.

At the same time, GCC 4.8.5 and GCC 8 both generate the exact same code before and after the problematic patch as Michael Ellerman pointed out. It might be worth pushing on a full revert in lieu of these workaround patches.

nickdesaulniers:
> Hmmm, thinking about this a little bit more, I think I like your patch in #195 a little bit more than mine because all of these functions have this code generation issue:

Shall I restart the CI of:
https://github.com/ClangBuiltLinux/continuous-integration/pull/195
then or do you want to pull that patch file into this PR?

> At the same time, GCC 4.8.5 and GCC 8 both generate the exact same code before and after the problematic patch as Michael Ellerman pointed out. It might be worth pushing on a full revert in lieu of these workaround patches.

We can suggest that "below the fold" when sending our patch.

nathanchance:
I'll push that patch here under the ppc32 directory and kick off a presubmit run.

nathanchance:
https://travis-ci.com/nathanchance/continuous-integration/builds/121137965

PR #201, opened 2019-08-20
Change android common cuttlefish configs to gki_defconfig

nickdesaulniers:
Nathan noticed that they were removed:
https://android.googlesource.com/kernel/common/+/3f0d9e29849d2ab72600269041fce18ff1b0b99a

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -79,7 +79,7 @@ setup_variables() {
 
     "arm64")
       case ${REPO} in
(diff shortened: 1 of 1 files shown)


nathanchance:
This will break `android-4.9` and `android-4.14` because there is no `gki_defconfig` in those branches; we'll need a versioning heuristic here.

PR #210, opened 2019-10-04
link more arm64 stable with LLD

nickdesaulniers:
Older stable is failing seemingly related to incremental linking being
removed in 4.17-rc1:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f49821ee32b76b1a356fab17316eb62430182ecf
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5e18f0290f2eccb0f64e87fbfca6395f11985b4b

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,8 +54,8 @@ matrix:
     - name: "ARCH=arm32_v7 REPO=4.19"
       env: ARCH=arm32_v7 REPO=4.19
       if: type = cron
-    - name: "ARCH=arm64 REPO=4.19"
-      env: ARCH=arm64 REPO=4.19
+    - name: "ARCH=arm64 LD=ld.lld REPO=4.19"
+      env: ARCH=arm64 LD=ld.lld-10 REPO=4.19
       if: type = cron
     - name: "ARCH=ppc64le REPO=4.19"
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
I'm not sure yet why older branches are failing precisely, might just run a bisection.

nathanchance:
Hmmm, 4.14 works fine for me. 4.9 fails but it appears fixed with:

https://github.com/torvalds/linux/commit/ce99d0bf312daf0178e640da9e3c93b773a67e7d
https://github.com/torvalds/linux/commit/3d6a7b99e3fa29b92d6288487e057e0a596bd2b0
https://github.com/torvalds/linux/commit/2893af07e507e993ad71ca6d66a7b02be741571c
https://github.com/torvalds/linux/commit/c931d34ea0853d41349e93f871bd3f17f1c03a6b

Backported series for 4.9.194 here (`curl -LSs <raw_link> | git am`): https://gist.github.com/97a392a12675cf4075ea20d19ac52566

There are several patches that need to be backported for `CONFIG_RANDOMIZE_BASE` to work properly (which I doubt would be possible because I know some of the patches were somewhat large and dependent on stuff in previous releases).

PR #211, opened 2019-10-04
enable llvm-size by default for all arch's

nickdesaulniers:
though it's only used on s390

### DIFF
--- a/usage.txt
+++ b/usage.txt
@@ -34,6 +34,11 @@ Environment variables:
         common-4.14
         common-4.9
         common-4.4
+  OBJSIZE:
+      If no OBJSIZE value is specified, the script will attempt to set OBJSIZE
+      to llvm-size-10 llvm-size-9, llvm-size-8, llvm-size-7, llvm-size, then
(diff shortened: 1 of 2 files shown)


nickdesaulniers:
probably should update usage.txt

nickdesaulniers:
done; yes, it's pointless, but it's one less thing in the Makefile for me to worry about.

usage.txt
```diff
@@ -34,6 +34,11 @@ Environment variables:
         common-4.14
         common-4.9
         common-4.4
+  OBJSIZE:
+      If no OBJSIZE value is specified, the script will attempt to set OBJSIZE
+      to llvm-OBJSIZE-10 llvm-OBJSIZE-9, llvm-OBJSIZE-8, llvm-OBJSIZE-7,
```
nathanchance: I think you meant for `llvm-size` here instead of `llvm-OBJSIZE`? Same with `${CROSS_COMPILE}size` instead of `${CROSS_COMPILE}OBJSIZE`?

nickdesaulniers: done

Issue #183, opened 2019-06-26
Add ability to specify more LLVM tools in driver.sh


tpimh:
Currently, the following can be specified:
- [X] AR
- [ ] AS (WIP PR #163, if set to clang, should also enable integrated-as)
- [X] CC
- [X] LD
- [X] NM
- [ ] OBJCOPY (was reverted: https://github.com/ClangBuiltLinux/continuous-integration/commit/684262f304e3f840a95eeeaef71c95a52f44fa3f)
- [X] OBJDUMP
- [ ] STRIP (was reverted: https://github.com/ClangBuiltLinux/continuous-integration/commit/e7afaa8d1c4dd6db5a26b8616d45cc2689b723e6)

Once it's done, work on #73 can be continued.

nickdesaulniers:
Re: OBJCOPY, https://github.com/ClangBuiltLinux/linux/issues/480 needs to be retested (I think it's fixed?)

nathanchance:
I am testing `llvm-objcopy` now: https://travis-ci.com/nathanchance/continuous-integration/builds/117031908

tpimh:
All tests show green, probably `llvm-objcopy` is good enough now (`llvm-strip` should be as well).

nathanchance:
Unless `llvm-strip` started supporting globs (it doesn't look like it according to the bug in https://github.com/ClangBuiltLinux/linux/issues/478, it will still be broken.

On that note, I forgot to test arm64 4.4, which was broken with `llvm-objcopy`: https://github.com/ClangBuiltLinux/linux/issues/478#issuecomment-492420162

nathanchance:
`llvm-objcopy` is still broken for arm64 4.4 because globs are not supported: https://travis-ci.com/nathanchance/continuous-integration/jobs/211920146

tapaswenipathak:
Hi folks: Didn't read much, is the ticket done?

tpimh:
`llvm-strip`, `llvm-objcopy` and clang's integrated as are still not used during the build process. No support for globs (https://github.com/ClangBuiltLinux/linux/issues/616) is a big deal for strip and objcopy for some targets.

tapaswenipathak:
Hey @tpimh! Cool, I see what item you are pointing out. Have we added/implemented contributor covenant code of conduct in the main repositories?

tpimh:
> Have we added/implemented contributor covenant code of conduct in the main repositories?

I don't think so.

[...]

Issue #164, opened 2019-05-14
Recent failures after turning on more LLVM tools


nathanchance:
* https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/199950114
  https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/199950118
  https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/199950120

  Stuff along the lines of:

  ```
  0000000000000000 R_AARCH64_ABS64 sort
  0000000000000008 R_AARCH64_ABS64 __ksymtab_strings
  drivers/firmware/efi/libstub/lib-sort.stub.o: absolute symbol references not allowed in the EFI stub
  ```

  It looks like this is probably caused by `llvm-strip` not supporting the globs in `STUBCOPY_RM-y`, which @rupprecht noted [here](https://github.com/ClangBuiltLinux/linux/issues/435#issuecomment-477387356). These were removed upstream in https://github.com/torvalds/linux/commit/02562d0ca1084a688ac5c92e0e92947f62f13093 because of https://github.com/torvalds/linux/commit/f922c4abdf7648523589abee9460c87f51630d2f. https://github.com/torvalds/linux/commit/f922c4abdf7648523589abee9460c87f51630d2f decently backports to 4.9 and 4.14, not so clean for 4.4. Might be wise to conditionally turn on `llvm-strip`.

* https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/199950119
  https://travis-ci.com/ClangBuiltLinux/continuous-integration/jobs/199950121

  ```
  llvm-objcopy-9: error: 'firmware/e100/d101m_ucode.bin.ihex': The file was not recognized as a valid object file
  ```

  Relevant bits of Makefile:

  ```
  fw-shipped-$(CONFIG_E100) += e100/d101m_ucode.bin e100/d101s_ucode.bin \
			     e100/d102e_ucode.bin
  ...
  quiet_cmd_ihex  = IHEX    $@
      cmd_ihex  = $(OBJCOPY) -Iihex -Obinary $< $@
  ```

  This was "resolved" upstream by https://github.com/torvalds/linux/commit/5620a0d1aacd554ebebcff373e31107bb1ef7769, which we obviously can't backport. We could apply a local patch that removes the `fw-shipped` rule in the Makefile or fix `llvm-objcopy` (if it is even broken).

nickdesaulniers:
[...]

PR #212, opened 2019-10-04
selectively re-enable llvm-objcopy

nickdesaulniers:
just for arm64, back to LTS 4.9.

4.4 is still broken:
https://github.com/ClangBuiltLinux/continuous-integration/issues/183#issuecomment-506861074

Previously reverted in 684262f304e3f840a95eeeaef71c95a52f44fa3f

https://github.com/ClangBuiltLinux/linux/issues/480
https://github.com/ClangBuiltLinux/continuous-integration/issues/164


tpimh:
What about llvm-strip? Does it still have problems on arm64? It makes sense to add both objcopy and strip in a single pull-request.

PR #213, opened 2019-10-04
driver: Fix arm32_v6 flakiness

nathanchance:
This build has been flaky since the 5.4 merge window it seems. I cannot
reproduce it locally so I believe it has something to do with Travis.
Increase the timeout to 4 minutes to make absolutely sure that the boot
has failed, rather than just being very long.

Presubmit: https://travis-ci.com/nathanchance/continuous-integration/builds/130365032

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -349,11 +349,14 @@ boot_qemu() {
   fi
 
   test -e ${kernel_image}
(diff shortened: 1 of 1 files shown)


nathanchance:
I believe it will but we will see over the next week or so :) thanks for the review!

PR #225, opened 2019-12-12
add 5.4, android-5.4, and android-mainline branches

nickdesaulniers:
Also, this is the first LTS branch for mips.  We could add some ppc,
too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,8 +51,8 @@ matrix:
       env: ARCH=x86_64 LD=ld.lld-10 REPO=linux-next
       if: type = cron
     # stable
-    - name: "ARCH=arm32_v7 REPO=5.4"
-      env: ARCH=arm32_v7 REPO=5.4
+    - name: "ARCH=arm32_v7 LD=ld.lld REPO=5.4"
+      env: ARCH=arm32_v7 LD=ld.lld REPO=5.4
       if: type = cron
     - name: "ARCH=arm64 LD=ld.lld REPO=5.4"
       env: ARCH=arm64 LD=ld.lld-10 REPO=5.4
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -51,6 +51,21 @@ matrix:
       env: ARCH=x86_64 LD=ld.lld-10 REPO=linux-next
       if: type = cron
     # stable
+    - name: "ARCH=arm32_v7 REPO=5.4"
```
nathanchance: This should be able to be linked with `ld.lld`.

.travis.yml
```diff
@@ -51,6 +51,21 @@ matrix:
...
+      env: ARCH=arm32_v7 REPO=5.4
+      if: type = cron
+    - name: "ARCH=arm64 LD=ld.lld REPO=5.4"
+      env: ARCH=arm64 LD=ld.lld-10 REPO=5.4
+      if: type = cron
+    - name: "ARCH=mipsel REPO=5.4"
+      env: ARCH=mipsel REPO=5.4
+      if: type = cron
+    - name: "ARCH=ppc64le REPO=5.4"
```
nathanchance: This is going to be broken until [c9029ef9c957](https://git.kernel.org/linus/c9029ef9c95765e7b63c4d9aa780674447db1ec0) ("powerpc: Avoid clang warnings around setjmp and longjmp") is backported to stable (it should be picked up soon because it was cc'd). I'd be fine with carrying out of tree but I don't think it is critical.

I think that ppc32 should work fine right now though.

.travis.yml
nickdesaulniers: done

.travis.yml
nickdesaulniers: done

.travis.yml
```diff
@@ -51,6 +51,21 @@ matrix:
       env: ARCH=x86_64 LD=ld.lld-10 REPO=linux-next
       if: type = cron
     # stable
+    - name: "ARCH=arm32_v7 LD=ld.lld REPO=5.4"
+      env: ARCH=arm32_v7 LD=ld.lld REPO=5.4
```
nathanchance: Needs to be `LD=ld.lld-10`.

nickdesaulniers: done

PR #228, opened 2020-02-07
Update travis configuration to pass new build config validation without warnings

tpimh:
Before this change travis config validation produces the following messages:

- `[WARN]` `root`: deprecated key `sudo` (The key `sudo` has no effect anymore.)
- `[INFO]` `root`: key `matrix` is an alias for `jobs`, using `jobs`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: cpp
-matrix:
+jobs:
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
thanks 🐊 !

Issue #152, opened 2019-04-24
s390 support


nickdesaulniers:
@arndb sent a lot of patches for 390 recently.  I think there's no more build errors?  If so, maybe there's a defconfig or certain machine we can start testing?

nathanchance:
I just tried building this in our Docker image. After installing `binutils-s390x-linux-gnu` and applying @arndb's patches from [the s390 tree](https://git.kernel.org/pub/scm/linux/kernel/git/s390/linux.git/log/?h=features) on top of `next-20190503`:

```
$ git diff c263a4e990b7^..ce968f6012f6 | git apply
$ make -j16 ARCH=s390 CROSS_COMPILE=s390x-linux-gnu- CC=clang-9 HOSTCC=clang-9 O=/out performance_defconfig bzImage
```

[Build log](https://gist.github.com/0eafed9755907f3bc6c35ca30fb19f18)

I don't know if I am holding wrong or what but the most glaring issues (can file separate issues for these if needed):

- [ ] `-Waddress-of-packed-member` isn't getting disabled properly (maybe something with `cc-option` as well because `-fno-null-pointer-checks` isn't being enabled either).

- [ ] Error in `lib/raid6/s390vx8.c` (maybe a binutils error?):

  ```
  lib/raid6/s390vx8.c:63:16: error: invalid operand in inline asm: 'VLM $2,$3,0,${1:r}'
          asm volatile ("VLM %2,%3,0,%r1"
                        ^
  lib/raid6/s390vx8.c:63:16: error: invalid operand in inline asm: 'VLM $2,$3,0,${1:r}'
  lib/raid6/s390vx8.c:63:16: error: invalid operand in inline asm: 'VLM $2,$3,0,${1:r}'
  lib/raid6/s390vx8.c:63:16: error: invalid operand in inline asm: 'VLM $2,$3,0,${1:r}'
  lib/raid6/s390vx8.c:63:16: error: invalid operand in inline asm: 'VLM $2,$3,0,${1:r}'
  lib/raid6/s390vx8.c:63:16: error: invalid operand in inline asm: 'VLM $2,$3,0,${1:r}'
  6 errors generated.
  ```

For CI, we'll need to get a proper set of QEMU flags and a rootfs going.

arndb:
On Fri, May 3, 2019 at 10:39 PM Nathan Chancellor <notifications@github.com>
wrote:

> I just tried building this in our Docker image. After installing
[...]

PR #229, opened 2020-02-07
S390 support

nathanchance:
Closes: #152 
Requires: https://github.com/ClangBuiltLinux/dockerimage/pull/40

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ jobs:
       env: ARCH=ppc64
     - name: "ARCH=ppc64le LD=ld.lld"
       env: ARCH=ppc64le LD=ld.lld-11
+    - name: "ARCH=s390"
+      env: ARCH=s390
     - name: "ARCH=x86_64 LD=ld.lld"
       env: ARCH=x86_64 LD=ld.lld-11
     # linux (cron only)
@@ -46,6 +48,9 @@ jobs:
       if: type = cron
     - name: "ARCH=ppc64le LD=ld.lld REPO=linux-next"
       env: ARCH=ppc64le LD=ld.lld-11 REPO=linux-next
+      if: type = cron
+    - name: "ARCH=s390 REPO=linux-next"
+      env: ARCH=s390 REPO=linux-next
       if: type = cron
     - name: "ARCH=x86_64 LD=ld.lld REPO=linux-next"
       env: ARCH=x86_64 LD=ld.lld-11 REPO=linux-next
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ jobs:
       env: ARCH=ppc64
     - name: "ARCH=ppc64le LD=ld.lld"
       env: ARCH=ppc64le LD=ld.lld-11
-    - name: "ARCH=s390"
+    - name: "ARCH=s390 BOOT=0"
       env: ARCH=s390
     - name: "ARCH=x86_64 LD=ld.lld"
       env: ARCH=x86_64 LD=ld.lld-11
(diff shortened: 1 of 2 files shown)


nickdesaulniers:
it doesn't boot? (we could reuse that qemu skip for hexagon (cc @androm3da )

nathanchance:
I did not try per the comment in Guenter’s build script: https://github.com/groeck/linux-build-test/blob/b8650ca8cdef921992b71db582b9bbdc4efb0ab3/rootfs/s390/run-qemu-s390.sh#L22

nickdesaulniers:
but it looks like @groeck still boot tests s390, he just does `defconfig`+`CONFIG_MARCH_Z900=y`, which is what we should do, too.

We already do this for a few arch's, a la: https://github.com/ClangBuiltLinux/continuous-integration/blob/fa15d669c53a70beebb9d612cc7c2f4ea42b5879/driver.sh#L332

nathanchance:
We can’t do that with clang because of https://git.kernel.org/linus/c263a4e990b7296b074e33aa077239a0a28a818e.

.travis.yml
```diff
@@ -18,6 +18,8 @@ jobs:
       env: ARCH=ppc64
     - name: "ARCH=ppc64le LD=ld.lld"
       env: ARCH=ppc64le LD=ld.lld-11
+    - name: "ARCH=s390"
```
nickdesaulniers: can you wire this up so that we explicitly put `BOOT=0` or something, such that it's explicit in the config that we don't boot test?  That will make it more obvious in the build results, and we won't have to make further changes to driver.sh when turning on builds of hexagon.

nathanchance: Sure, done.

driver.sh
```diff
@@ -139,6 +139,13 @@ setup_variables() {
       export ARCH=powerpc
       export CROSS_COMPILE=powerpc64le-linux-gnu- ;;
 
+    "s390")
+        config=defconfig
+        image_name=bzImage
+        using_qemu=false
```
nickdesaulniers: right but i meant removing this

nathanchance: If we remove that from `driver.sh`, it requires a local user to explicitly add the `BOOT=0`, otherwise the build won’t even start because QEMU won’t be found.

tpimh:
Instead of db908a6 we should add [qemu-system-misc](https://packages.debian.org/buster/otherosfs/qemu-system-misc) package to dockerimage, create a rootfs image (need some research, s390 support seems to be removed from buildroot: https://github.com/buildroot/buildroot/commit/b24c3215c13b0f25979c744af4b2e58359b5723b) and run it like this:
`qemu-system-s390x -m 512m -machine s390-ccw-virtio -initrd images/s390/rootfs.cpio -display none -serial mon:stdio -kernel linux/arch/s390/boot/bzImage`

tpimh:
I wonder if we can use initrd from [moonbuggy image from the 2014 QEMU Advent Calendar](https://www.qemu-advent-calendar.org/2014/download/s390-moon-buggy.tar.xz)

tpimh:
Just checked the moonbuggy image, it's not very useful: only a couple shared libraries and a game binary as init. I'm now checking out [gentoo netboot initramfs for s390x](http://distfiles.gentoo.org/releases/s390/autobuilds/current-netboot-s390x-initramfs/netboot-s390x-initramfs-20191211T084703Z), it has busybox in it.

tpimh:
Just built the kernel, tried to boot with QEMU, got Kernel fault:
```
KASLR disabled: CPU has no PRNG
Linux version 5.5.0-gf75716570 (driver@clangbuiltlinux) #1 SMP Thu Jan 1 00:00:00 UTC 1970Kernel fault: interruption code 0005 ilc:2
PSW : 0000200180000000 00000000000175d4
      R:0 T:0 IO:0 EX:0 Key:0 M:0 W:0 P:0 AS:0 CC:2 PM:0 RI:0 EA:3
GPRS: 00000000fffffff8 0000000c00000009 00000003fffffff4 0000000000000008
      0000000000000010 00000000fffffff0 00000000fffffff4 0000000000000004
      000000000000000c 0000000000000000 00000000fffffffc 00000000009495a8
      0000000000000002 0000000000000001 0000000000000000 000000000000bb58
```
At least the console is functional.

PR #231, opened 2020-02-08
driver.sh: Use the '-s' flag for make to show only warnings/errors

nathanchance:


nathanchance:
Actually, this could cause Travis to timeout if we go 10 minutes without a warning: https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out

tpimh:
Also if there are multiple errors, it's easier to see where output from one compilation command ends and next output begins without `-s`. With `-s` it's just a long output without any breaks.

nathanchance:
Yeah, as you can see from presubmit, this won’t work :(

PR #234, opened 2020-03-10
add dtc patch for yylloc

nickdesaulniers:
https://git.kernel.org/pub/scm/utils/dtc/dtc.git/commit/?id=0e9225eb0dfec51def612b928d2f1836b092bc7e

Reported-by: Nathan Chancellor <natechancellor@gmail.com>


nathanchance:
The patch needs to be in `patches/llvm-all/kernel-all`.

https://github.com/ClangBuiltLinux/continuous-integration/blob/a85e3e44c2570847e22ad8f92f317c2b007c4517/driver.sh#L330

The patch applies cleanly to 4.4 stable for me locally but might not be a bad idea to get some presubmit on some of the other older trees.

nickdesaulniers:
(ready for rereview)

PR #238, opened 2020-03-18
Build RISC-V target only for -next

tpimh:
As suggested in https://github.com/ClangBuiltLinux/continuous-integration/pull/70#issuecomment-599714608, RISC-V targets are only enabled on linux-next, so they would build fine without the patches. Build logs for these targets can be found [here](https://github.com/ClangBuiltLinux/continuous-integration/runs/512210549).

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -149,6 +149,12 @@ setup_variables() {
       export ARCH=powerpc
       export CROSS_COMPILE=powerpc64le-linux-gnu- ;;
 
+    "riscv")
+      config=defconfig
+      image_name=vmlinux
+      using_qemu=false
+      export CROSS_COMPILE=riscv64-linux-gnu- ;;
+
     "s390")
         config=defconfig
         image_name=bzImage
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,9 @@ jobs:
     - name: "ARCH=ppc64le LD=ld.lld REPO=linux-next"
       env: ARCH=ppc64le LD=ld.lld-11 REPO=linux-next
       if: type = cron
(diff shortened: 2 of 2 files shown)


driver.sh
```diff
@@ -149,6 +149,20 @@ setup_variables() {
       export ARCH=powerpc
       export CROSS_COMPILE=powerpc64le-linux-gnu- ;;
 
+    "riscv32")
+      config=defconfig
+      image_name=vmlinux
+      using_qemu=false
+      export ARCH=riscv
+      export CROSS_COMPILE=riscv64-linux-gnu- ;;
```
nathanchance: Should this be `riscv32-linux-gnu-`?

driver.sh
```diff
@@ -149,6 +149,20 @@ setup_variables() {
       export ARCH=powerpc
       export CROSS_COMPILE=powerpc64le-linux-gnu- ;;
 
+    "riscv32")
+      config=defconfig
```
nathanchance: Should this be `rv32_defconfig`?

driver.sh
tpimh: I don't think debian even has riscv32 binutils in the repo.

driver.sh
tpimh: Probably these two issues are the reason why the kernels produced are nearly identical: https://github.com/ClangBuiltLinux/continuous-integration/pull/70#issuecomment-599708751

driver.sh
nathanchance: Doesn't look like it. Seems like we should only be building for riscv64 then.

nickdesaulniers: @palmer-dabbelt , any recommendations on 32b riscv binutils?

Issue #239, opened 2020-03-18
arm64 4.4 errors


nickdesaulniers:
last nights arm64 4.4 build failed https://travis-ci.com/github/ClangBuiltLinux/continuous-integration/jobs/299396126?utm_medium=notification&utm_source=email due to multiple definitions of `yylloc`.  What's weird is that:
1. it's picking up the patch :
```
+ git apply -v -3 ../patches/llvm-11/4.4/arch-all/0001-scripts-dtc-Remove-redundant-YYLOC-global-declaratio.patch
Checking patch scripts/dtc/dtc-lexer.l...
Hunk #1 succeeded at 38 (offset 15 lines).
Applied patch scripts/dtc/dtc-lexer.l cleanly.
```
2. I can't reproduce locally in stable linux-4.4.y

ah but looks like I can via this repo `CC=clang ARCH=arm64 REPO=4.4 ./driver.sh`, hmmm

this diff seems to work for me:
```diff
diff --git a/scripts/dtc/dtc-lexer.l b/scripts/dtc/dtc-lexer.l
index 0ee1caf03dd0..a9b58b71e25e 100644
--- a/scripts/dtc/dtc-lexer.l
+++ b/scripts/dtc/dtc-lexer.l
@@ -38,7 +38,6 @@ LINECOMMENT   "//".*\n
 #include "srcpos.h"
 #include "dtc-parser.tab.h"
 
-YYLTYPE yylloc;
 extern bool treesource_error;
 
 /* CAUTION: this will stop working if we ever use yyless() or yyunput() */
diff --git a/scripts/dtc/dtc-lexer.lex.c_shipped b/scripts/dtc/dtc-lexer.lex.c_shipped
index 11cd78e72305..26932b00cf2d 100644
--- a/scripts/dtc/dtc-lexer.lex.c_shipped
+++ b/scripts/dtc/dtc-lexer.lex.c_shipped
@@ -637,7 +637,6 @@ char *yytext;
 #include "srcpos.h"
 #include "dtc-parser.tab.h"
 
-YYLTYPE yylloc;
 extern bool treesource_error;
 
 /* CAUTION: this will stop working if we ever use yyless() or yyunput() */
```

nathanchance:
Hmmm I thought I fixed this but I missed 4.4.

4.9 and 4.14 suffer from the same issue but I got those in #235. Copying the patch that is in the 4.9/4.14 folders should fix this.

PR #240, opened 2020-03-18
fix arm64 4.4 build

nickdesaulniers:
`$ cp patches/llvm-11/4.9/arch-all/0001-scripts-dtc-Remove-redundant-YYLOC-global-declaratio.patch patches/llvm-11/4.4/arch-all/0001-scripts-dtc-Remove-redundant-YYLOC-global-declaratio.patch`

Fixes: #239


nickdesaulniers:
tested locally, merging

PR #242, opened 2020-03-26
Add machinery for AS=clang builds

nickdesaulniers:


.travis.yml
```diff
@@ -22,8 +22,8 @@ jobs:
       env: ARCH=ppc64le LD=ld.lld-11
     - name: "ARCH=s390 BOOT=0"
       env: ARCH=s390
-    - name: "ARCH=x86_64 LD=ld.lld"
-      env: ARCH=x86_64 LD=ld.lld-11
+    - name: "ARCH=x86_64 LD=ld.lld AS=clang"
+      env: ARCH=x86_64 LD=ld.lld-11 AS=clang
```
nathanchance: Needs to be `clang-11` due to https://apt.llvm.org binaries.

Follow up thoughts?

1. Swap `AS=clang` with `LD=ld.lld-11` to keep it alphabetical?

2. Should we do a GNU as build on cron?

patches/llvm-11/linux/arch-all/0001-elfnote-mark-all-.note-sections-SHF_ALLOC.patch
```diff
@@ -0,0 +1,53 @@
+From 2a32e42dae19146939a951c2368bfeb0b20e0698 Mon Sep 17 00:00:00 2001
```
nathanchance: 1. While this is a patch that all architectures will need to work with the integrated assembler, I'd rather add it on a architecture by architecture basis so that when the patch gets added, every single build within the version we are testing does not go red. Personal opinion though, not blocking.

2. If we are adding mainline, we should add -next to hunt down regressions quicker.

nickdesaulniers: We can't add -next; https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?h=core/objtool&id=644592d328370af4b3e027b7b1ae9f81613782d8 means that https://github.com/ClangBuiltLinux/linux/issues/669 becomes an error.

.travis.yml
nickdesaulniers: done.

patches/llvm-11/linux/arch-all/0001-elfnote-mark-all-.note-sections-SHF_ALLOC.patch
nickdesaulniers: Otherwise, done. Ready for rereview.

nathanchance: Hmmm... then adding this target seems kind of pointless since we probably won't even get a week of being green; 5.7 is likely to be released on Sunday and the tip pull requests tend to be the first to go out and get merged, meaning we'd get from now until probably Tuesday of green.

Maybe we should ship one of your objtool patches as well? https://github.com/ClangBuiltLinux/linux/issues/669#issuecomment-604132262

nickdesaulniers: > we probably won't even get a week of being green

Lol, reminds me of this passage:
> "But it won't have amounted to anything," said Grego,
> "If your children die, then it was all a waste."
> "No," said Olhado quietly.
> ...
> Grego shook his head. "Yes it does, Olhado. Death undoes
> everything."
> Olhado shrugged, "Then why do you bother doing
> everything, Grego? Because someday you will die. Why
> should anyone have children? Someday they will die,
> their children will die, all children will die.  Someday stars
> will wind down or blow up. Someday death will cover us
> all like the water of a lake and perhaps nothing will ever
> come to the surface to show that we were ever there. But
> we *were* there, and during the time we lived, we were alive.
> That's the truth - what is, what was, what will be - not
> what could be, what should have been, what never can be.
> If we die, then our death has meaning to the rest of the
> universe. Even if our lives are unknown, the fact that someone
> lived here, and died, that will have repercussions, that
> will shape the universe."
> Children of the Mind, Orson Scott Card

But yeah existential crises aside, how about I drop the builds and patch, and just keep the modifications to driver.sh and usage.txt, that way the machinery is in place for when we no longer need to carry any patches?

nathanchance: > Lol, reminds me of this passage:

Hah good one :)

> But yeah existential crises aside, how about I drop the builds and patch, and just keep the modifications to driver.sh and usage.txt, that way the machinery is in place for when we no longer need to carry any patches?

Sounds good to me.

nickdesaulniers: done.

nickdesaulniers:
rebased

nickdesaulniers:
Wait, don't I need to export this in .travis.yml via `--env AS=${AS}`?

nathanchance:
> Wait, don't I need to export this in .travis.yml via `--env AS=${AS}`?

Whoops, yes!

nickdesaulniers:
done

Issue #247, opened 2020-04-01
move versioned builds to PATH modifications, rather than version numbers


nickdesaulniers:
as discussed on https://groups.google.com/g/clang-built-linux/c/1j6_AfP0CMQ with @masahir0y and @nathanchance , debian's `clang-9` versioned binaries are symlinks to versioned file paths to unsuffixed binaries.  We should modify driver.sh to export this `$PATH` and use `CC=clang` (unsuffixed) rather than `CC=clang-9`.

PR #250, opened 2020-04-01
ci: Get rid of versioning binary scheme

nathanchance:
Debian symlinks their versioned binaries (clang-11, ld.lld-11, etc) in
/usr/bin from /usr/lib/llvm-#/bin. As a result, we do not need to
account for the suffixes, we can just add that folder to PATH if it
exists.

This simplies our CI matrix and handling within the script.

The CI failure in presubmit below is related to
https://github.com/ClangBuiltLinux/linux/issues/962, I just wanted to
show that the versioning detection still works properly afte this
patch.

Closes: https://github.com/ClangBuiltLinux/continuous-integration/issues/247
Presubmit: https://travis-ci.com/github/nathanchance/continuous-integration/builds/157635172

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -151,11 +151,6 @@ function update_boot_utils() {
   fi
 }
 
-# Generates a list of binary versions based on latest_llvm_version and oldest_llvm_version
-# Example: gen_bin_list clang spits out clang-10 clang-9 clang-8...
-gen_bin_list() {
-    seq -f "${1:?}-%.0f" "${latest_llvm_version}" -1 "${oldest_llvm_version}"
-}
 
 check_dependencies() {
   # Check for existence of needed binaries
@@ -170,9 +165,16 @@ check_dependencies() {
   oldest_llvm_version=7
   latest_llvm_version=$(curl -LSs https://raw.githubusercontent.com/llvm/llvm-project/master/llvm/CMakeLists.txt | grep -s -F "set(LLVM_VERSION_MAJOR" | cut -d ' ' -f 4 | sed 's/)//')
 
-  for readelf in $(gen_bin_list llvm-readelf) llvm-readelf; do
-    command -v ${readelf} &>/dev/null && break
+  for llvm_version in $(seq "${latest_llvm_version}" -1 "${oldest_llvm_version}"); do
+    debian_llvm_bin=/usr/lib/llvm-${llvm_version}/bin
+    if [[ -d ${debian_llvm_bin} ]]; then
+      export PATH=${debian_llvm_bin}:${PATH}
+      break
+    fi
   done
+
+  READELF=llvm-readelf
+  command -v "${READELF}"
 
   # Check for LD, CC, and AR environmental variables
   # and print the version string of each. If CC and AR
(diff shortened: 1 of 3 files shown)


driver.sh
```diff
@@ -170,10 +165,17 @@ check_dependencies() {
   oldest_llvm_version=7
   latest_llvm_version=$(curl -LSs https://raw.githubusercontent.com/llvm/llvm-project/master/llvm/CMakeLists.txt | grep -s -F "set(LLVM_VERSION_MAJOR" | cut -d ' ' -f 4 | sed 's/)//')
 
-  for readelf in $(gen_bin_list llvm-readelf) llvm-readelf; do
-    command -v ${readelf} &>/dev/null && break
+  for llvm_version in $(seq "${latest_llvm_version}" -1 "${oldest_llvm_version}"); do
+    debian_llvm_bin=/usr/lib/llvm-${llvm_version}/bin
+    if [[ -d ${debian_llvm_bin} ]]; then
+      export PATH=${debian_llvm_bin}:${PATH}
+      break
```
nickdesaulniers, line 172: else die

driver.sh
```diff
@@ -185,9 +187,8 @@ check_dependencies() {
   "${AS:="${CROSS_COMPILE:-}"as}" --version
 
   if [[ -z "${CC:-}" ]]; then
```
nickdesaulniers, line 189: wont this ALWAYS be true?

driver.sh
```diff
@@ -207,28 +208,28 @@ check_dependencies() {
...
     done
   fi
 
   if [[ -z "${OBJSIZE:-}" ]]; then
-    for OBJSIZE in $(gen_bin_list llvm-size) llvm-size "${CROSS_COMPILE:-}"size; do
-      command -v ${OBJSIZE} 2>/dev/null && break
+    for OBJSIZE in llvm-size "${CROSS_COMPILE:-}"size; do
+      command -v "${OBJSIZE}" 2>/dev/null && break
     done
   fi
```
nickdesaulniers, line 234: do you think that you could write a function that accepted `OBJSIZE` or `OBJDUMP` or ... and did this, in order to DRY up this code?

driver.sh
nathanchance: If I understand correctly, you mean something like this?

```diff
diff --git a/driver.sh b/driver.sh
index 3e73efa..d617bfe 100755
--- a/driver.sh
+++ b/driver.sh
@@ -170,6 +170,9 @@ check_dependencies() {
     if [[ -d ${debian_llvm_bin} ]]; then
       export PATH=${debian_llvm_bin}:${PATH}
       break
+    else
+      echo "Can't find ${debian_llvm_bin}"
+      exit 1
     fi
   done

```

If so, that won't work.

1. We're iterating in a loop.

2. The user might not be on Debian/Ubuntu.

We'll catch these tools not being available down below though.

driver.sh
nathanchance: I'll see if I can think of something but

```
VAR=OBJDUMP
echo ${${VAR}}
```

to try and get `${OBJCOPY}` does not work :(

driver.sh
nathanchance: Not if a user passes in `CC=...`

driver.sh
nickdesaulniers: The names of the binaries are also different. Ok maybe not as easy to DRY up as I thought.

driver.sh
nickdesaulniers: `CC="ccache clang"`
got it

PR #251, opened 2020-04-02
travis: Use integrated assembler for RISCV

nathanchance:
This works fine and avoids the error in the issue below.

Closes: https://github.com/ClangBuiltLinux/linux/issues/967
Presubmit: https://travis-ci.com/github/nathanchance/continuous-integration/builds/157639008


nickdesaulniers:
Huh, I remember @palmer-dabbelt using the integrated assembler; I guess I forgot to check for that when this went in.  Easy fix now with #242 .  Also, ironic that the first architecture for which we enable AS doesn't actually boot yet...

PR #252, opened 2020-04-02
github: Add Actions workflow

nathanchance:
To ensure that any contributions are consistent and proper, add a GitHub
Action that runs shellcheck on any scripts.

For an example of how an error manifests and gets reported:

https://github.com/nathanchance/boot-utils/actions/runs/68702038


nathanchance:
Correct!

Issue #249, opened 2020-04-01
enable llvm-objcopy llvm-strip


nickdesaulniers:
just filing a bug, maybe it's a dup, but need to track this

PR #253, opened 2020-04-02
driver: Enable llvm-objcopy and llvm-strip

nathanchance:
except for s390, which does not appear to have any support for it in
the LLVM tools.

Tested locally with build-all.sh.

Closes: https://github.com/ClangBuiltLinux/continuous-integration/pull/212
Closes: https://github.com/ClangBuiltLinux/continuous-integration/issues/249

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -113,8 +113,12 @@ setup_variables() {
       config=defconfig
       make_target=bzImage
       using_qemu=false
-      OBJDUMP=s390x-linux-gnu-objdump
-      export CROSS_COMPILE=s390x-linux-gnu- ;;
+      export CROSS_COMPILE=s390x-linux-gnu-
+
+      # llvm-objcopy: error: invalid output format: 'elf64-s390'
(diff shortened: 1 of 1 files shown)


tpimh:
For s390 you set `OBJDUMP` variable, not `STRIP`. Is it intentional?

nathanchance:
Yes and no. I intentionally left it out because it does not appear to be used during a build (I have not seen an error). I can add it if you feel it necessary.

tpimh:
If it's not used during the build, no need to add it. I was just concerned because this pull request doesn't mention objdump anywhere else.

nathanchance:
Ah, gotcha. The objdump one is https://github.com/ClangBuiltLinux/linux/issues/859 and it has been there since we added s390 support.

nathanchance:
> cool, so we'll still need to turn this on in the travis yaml?

Nope, it is enabled by default now.

> Basically, it's a vote against supporting older versions.

Sure, I can push a patch after I merge this.

nickdesaulniers:
> Nope, it is enabled by default now.

*rereads code*

got it, just not s390, I see.

PR #255, opened 2020-04-02
driver: Use llvm-readelf

nathanchance:
Tested locally with build-all.sh


nathanchance:
Ack. Realistically, we should be able to backport the `LLVM=1` change without any hassle too.

PR #256, opened 2020-04-02
patches: Remove dtc patch

nathanchance:
NOTE: Do not merge this right away as the Android merges are still pending. If they aren't merged by the end of the afternoon West Coast time, I will just remove that commit from this pull.


nathanchance:
This is good to go now, all merges have happened in kernel/common.

Issue #257, opened 2020-04-02
disable emails for failed pulled requests


nickdesaulniers:
not sure if this is possible, but filing an issue to track. The build failure emails from travis are annoying for pull requests that fail to build.  I'd rather only receive email interruptions from the failed cron jobs.  Not sure if it's possible to limit those.

PR #258, opened 2020-04-02
travis: Send failure emails for crons only

nathanchance:
Let's see if this works.

Closes: #257


nathanchance:
Looks like it worked since the build failed as we expect it to but I did not get an email.

PR #266, opened 2020-04-16
travis: Switch to Ubuntu Docker image

nathanchance:
Depends on https://github.com/ClangBuiltLinux/dockerimage/pull/46.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -202,7 +202,7 @@ script:
         --rm \
         --workdir /travis \
         --volume ${TRAVIS_BUILD_DIR}:/travis \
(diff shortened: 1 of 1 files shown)


nathanchance:
This passes `./build-all.sh` for me (although there is an upstream s390 regression in -next: )https://lore.kernel.org/lkml/568AA084-BCC7-4638-AE65-DA64E547BA77@lca.pw/

nickdesaulniers:
yes, I noticed the s390 failure yesterday on our current docker image, so it's definitely not related to this new docker image.  Let's keep an eye on it, if it doesn't get cleared by Monday let's poke them.

nathanchance:
Looks like the s390 thing should be taken care of: https://lore.kernel.org/linux-next/20200417092725.4876a832@coco.lan/

PR #269, opened 2020-04-28
driver: Move to GNU objcopy for arm32_v5 for now

nathanchance:
There appears to be a bug in llvm-objcopy that was exposed by increasing
the page size to 64K for 32-bit ARM.

Move to GNU objcopy until that it fixed.

Presubmit: https://travis-ci.com/github/nathanchance/continuous-integration/builds/162491349

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -55,6 +55,9 @@ setup_variables() {
             make_target=zImage
             export ARCH=arm
             export CROSS_COMPILE=arm-linux-gnueabi-
+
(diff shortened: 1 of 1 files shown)


nathanchance:
> thanks, hopefully we can get this fixed in time for clang-10.1 release.

We should probably change the LLVM bug to objcopy and cc those developers for them to take a look.

nickdesaulniers:
> We should probably change the LLVM bug to objcopy and cc those developers for them to take a look.

Done.

PR #272, opened 2020-05-07
Allow objcopy and strip binaries to be specified

tpimh:
https://github.com/ClangBuiltLinux/continuous-integration/pull/253 added support for `objcopy` and `strip` utilities, but they couldn't be specified on the command line.

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -5,7 +5,7 @@ set -eu
 setup_variables() {
     while [[ ${#} -ge 1 ]]; do
         case ${1} in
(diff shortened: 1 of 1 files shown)


tpimh:
Or should `STRIP` be renamed to `OBJSTRIP`?

msfjarvis:
> Or should `STRIP` be renamed to `OBJSTRIP`?

`STRIP` looks like the variable used at least in the root `Makefile`.

nickdesaulniers:
🌋 🕋

PR #274, opened 2020-05-10
Enable integrated-as on selected arches on linux-next

tpimh:
As seen [here](https://travis-ci.com/github/ClangBuiltLinux/continuous-integration/builds/164981592), arm32_v5, arm64, x86 and x86_64 already have all necessary patches on linux-next to be built with integrated-as enabled. ARM64 still produces a lot of warnings due to https://github.com/ClangBuiltLinux/linux/issues/716.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ jobs:
     # linux (cron only)
     #
     # linux-next (cron only)
-    - name: "ARCH=arm32_v5 LD=ld.lld REPO=linux-next"
-      env: ARCH=arm32_v5 LD=ld.lld REPO=linux-next
+    - name: "ARCH=arm32_v5 LD=ld.lld LLVM_IAS=1 REPO=linux-next"
+      env: ARCH=arm32_v5 LD=ld.lld LLVM_IAS=1 REPO=linux-next
       if: type = cron
     - name: "ARCH=arm32_v6 REPO=linux-next"
       env: ARCH=arm32_v6 REPO=linux-next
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -31,17 +31,17 @@ jobs:
...
     - name: "ARCH=arm32_v6 REPO=linux-next"
       env: ARCH=arm32_v6 REPO=linux-next
       if: type = cron
     - name: "ARCH=arm32_v7 LD=ld.lld REPO=linux-next"
       env: ARCH=arm32_v7 LD=ld.lld REPO=linux-next
       if: type = cron
-    - name: "ARCH=arm64 LD=ld.lld REPO=linux-next"
-      env: ARCH=arm64 LD=ld.lld REPO=linux-next
+    - name: "ARCH=arm64 LD=ld.lld LLVM_IAS=1 REPO=linux-next"
+      env: ARCH=arm64 LD=ld.lld LLVM_IAS=1 REPO=linux-next
```
nickdesaulniers, line 44: I'm not comfortable enabling this for arm64 yet.  Maybe we can not add `-Wa,-gdwarf-2` when `LLVM_IAS=1` in the top level Makefile?

patches/llvm-all/linux-next/arm64/silence-dwarf2-warnings.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/Makefile b/Makefile
+index 9f9015664..ec3be353c 100644
+--- a/Makefile
++++ b/Makefile
+@@ -810,7 +810,6 @@ DEBUG_CFLAGS	+= -gsplit-dwarf
+ else
+ DEBUG_CFLAGS	+= -g
+ endif
```
nickdesaulniers, line 8: this should be an `ifndef` with `LLVM_IAS` (probably pretty close to the other reference to `LLVM_IAS` in this file) rather than deleting it.

nickdesaulniers:
thanks for the patch!

PR #278, opened 2020-06-07
travis: Move MIPS big endian to ld.bfd on mainline for now

nathanchance:
See commit 8cdfba2 ("travis: Move MIPS big endian to ld.bfd on -next
for now") for more reasoning.


nickdesaulniers:
thanks 🥥 ‼️

PR #281, opened 2020-06-16
Riscv boot

nickdesaulniers:
https://github.com/ClangBuiltLinux/linux/issues/867

### DIFF
--- a/driver.sh
+++ b/driver.sh
@@ -124,14 +124,12 @@ setup_variables() {
         "riscv")
             config=defconfig
             make_target=vmlinux
-            using_qemu=false
             export CROSS_COMPILE=riscv64-linux-gnu-
             ;;
 
         "s390")
             config=defconfig
             make_target=bzImage
-            using_qemu=false
             export CROSS_COMPILE=s390x-linux-gnu-
 
             # llvm-objcopy: error: invalid output format: 'elf64-s390'
@@ -416,7 +414,7 @@ build_linux() {
 }
 
 boot_qemu() {
-    ${using_qemu:=true} || return 0
+    if [[ "${BOOT}" == "0" ]]; then return 0; fi
     ./boot-utils/boot-qemu.sh -a "${SUBARCH}" -k "${tree}" -t "${timeout:-2}"m
 }
 
--- a/driver.sh
+++ b/driver.sh
@@ -128,6 +128,7 @@ setup_variables() {
             ;;
 
         "s390")
+            BOOT=false
             config=defconfig
             make_target=bzImage
             export CROSS_COMPILE=s390x-linux-gnu-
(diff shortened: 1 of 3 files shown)


driver.sh
```diff
@@ -416,7 +414,7 @@ build_linux() {
 }
 
 boot_qemu() {
-    ${using_qemu:=true} || return 0
+    if [[ "${BOOT}" == "0" ]]; then return 0; fi
```
nickdesaulniers, line 417: surely there's a better way to check this, but I guess you could set BOOT=1 to imply you want a boot.

nathanchance: @Nathan-Huckleberry 's patch is currently queued in the fixes branch which means it should be getting to Linus soon. We could just wait until that hits mainline to do this to avoid complicating things (since then we just remove `BOOT=0` from Travis and `using_qemu=false`.

However, if we want to go ahead with this:

I would turn `BOOT` into a boolean so that this statement stays the same but with a different name: `${BOOT:=true} || return 0`. It isn't like `BOOT=2` would have any meaning.

Additionally, `.travis.yml` needs to be updated to pass `BOOT` to the environmment and I think that `BOOT=false` should be set in the `s390` block to avoid issues with running that locally (and forgetting to see `BOOT=false`).

nickdesaulniers: > We could just wait until that hits mainline to do this to avoid complicating things

I prefer to have finer grain control here, and to have boot tests enabled ASAP.  It certainly is more complicated to only do boot tests for SOME branches of stable, and would like to avoid that if possible, since building w/o booting is significantly less valuable to me.

> It isn't like BOOT=2 would have any meaning.

Neither does `LLVM=2` 😛 But let me turn `BOOT` into a boolean.

nickdesaulniers: > .travis.yml needs to be updated to pass BOOT to the environmment

it does in this PR

nickdesaulniers: > I think that BOOT=false should be set in the s390 block to avoid issues with running that locally (and forgetting to see BOOT=false).

That I'm less excited about.  What if you forget that locally for riscv+mainline (won't matter once @Nathan-Huckleberry 's patch lands)?  Added it anyways.

nathanchance: > > We could just wait until that hits mainline to do this to avoid complicating things
> 
> I prefer to have finer grain control here, and to have boot tests enabled ASAP. It certainly is more complicated to only do boot tests for SOME branches of stable, and would like to avoid that if possible, since building w/o booting is significantly less valuable to me.

That's fair and convert the local `using_qemu` to a global `BOOT` variable is less complicated than I initially thought it would be.

> > It isn't like BOOT=2 would have any meaning.
> 
> Neither does `LLVM=2` 😛 But let me turn `BOOT` into a boolean.

Fair :P

> > I think that BOOT=false should be set in the s390 block to avoid issues with running that locally (and forgetting to see BOOT=false).
> 
> That I'm less excited about. What if you forget that locally for riscv+mainline (won't matter once @Nathan-Huckleberry 's patch lands)? Added it anyways.

We could move that logic in to `driver.sh` (like using a case statement on `REPO` or whatever) but I don't think it is strictly necessary. Ideally, that patch will be in mainline by `-rc2`, maybe `-rc3` at the latest.

PR #286, opened 2020-07-07
travis: Disable the integrated assembler for arm64

nathanchance:
Note, as of next-20200707, arm32 and arm64 boot is broken but should be
fixed in next-20200708 for reasons unrelated to LLVM:

https://lore.kernel.org/dmaengine/5f036d83.1c69fb81.10199.06d0@mx.google.com/
https://lore.kernel.org/dmaengine/159404871194.45151.3076873396834992441.stgit@djiang5-desk3.ch.intel.com/

No presubmit testing is done for that reason but this has been verified
locally.

Fixes: https://travis-ci.com/github/ClangBuiltLinux/continuous-integration/jobs/358277535
Link: https://github.com/ClangBuiltLinux/linux/issues/1078

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ jobs:
     - name: "ARCH=arm32_v7 LD=ld.lld REPO=linux-next"
       env: ARCH=arm32_v7 LD=ld.lld REPO=linux-next
       if: type = cron
-    - name: "ARCH=arm64 LD=ld.lld LLVM_IAS=1 REPO=linux-next"
-      env: ARCH=arm64 LD=ld.lld LLVM_IAS=1 REPO=linux-next
+    - name: "ARCH=arm64 LD=ld.lld REPO=linux-next"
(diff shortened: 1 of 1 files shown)


paulmenzel:
The mentioned

> Link: ClangBuiltLinux/linux#1081

seems to be incorrect, as that has nothing to do with arm64.

nathanchance:
Good catch, thanks!

samitolvanen:
Are the LKML links correct? How does the dmaengine patch break IAS?

nathanchance:
Sorry if I wasn't clear but the links are for a separate issue that at the time that I wrote that commit message would have prevented a positive presubmit run so I did not do one. That issue is fixed as of next-20200708 so I have started a presubmit run [here](https://travis-ci.com/github/nathanchance/continuous-integration/jobs/359158528) and I will update the commit message to remove that note and add that link when it is done.

PR #287, opened 2020-07-10
ci: Update certain Android branches for recent upstream changes

nathanchance:
android-4.19 has been moved to deprecated/android-4.19, meaning it will
no longer see updates. android-4.19-stable should be the equivalent
(similar to android-4.14-stable):

https://android-review.googlesource.com/c/kernel/manifest/+/1359844

Fixes the following builds:

https://travis-ci.com/github/ClangBuiltLinux/continuous-integration/jobs/359204121
https://travis-ci.com/github/ClangBuiltLinux/continuous-integration/jobs/359204126

Additionally, android-5.4 has been shuffled to android12-5.4:

https://android-review.googlesource.com/c/kernel/manifest/+/1358083
https://android-review.googlesource.com/c/kernel/manifest/+/1358088

android-5.4 still technically exists but do not rely on that always
being true and just do that change now so that we do not have to do it
later when the build breaks.

Presubmit: https://travis-ci.com/github/nathanchance/continuous-integration/builds/175084826

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,11 +135,11 @@ jobs:
     - name: "ARCH=arm64 REPO=android-4.14-stable"
       env: ARCH=arm64 REPO=android-4.14-stable
       if: type = cron
-    - name: "ARCH=arm64 REPO=android-4.19"
-      env: ARCH=arm64 REPO=android-4.19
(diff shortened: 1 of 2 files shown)


nickdesaulniers:
thanks! 🎆 I had some heads up about this this week, but not on when the final decision was made.

PR #290, opened 2020-07-27
patches: llvm-12: Remove linux patch

nathanchance:
It has been merged into mainline:

https://git.kernel.org/linus/ca9b31f6bb9c6aa9b4e5f0792f39a97bbffb8c51

PR #291, opened 2020-07-27
Two fixes for PowerPC and RISC-V failures

nathanchance:
These two commits along with #290 should make CI go green again.


nathanchance:
The RISC-V one does, I linked it in that commit message. The PowerPC one is fixed, I will try to remember to turn the tests back on if QEMU gets uprev'd. Disabling the x86 torture tests came from @groeck's boot tests where I initially took the configs, I am not sure that is a clang specific issue.

PR #301, opened 2020-08-20
llvm=1

nickdesaulniers:


nickdesaulniers:
LGTM: https://travis-ci.com/github/ClangBuiltLinux/continuous-integration/jobs/375549925#L521

driver.sh
```diff
@@ -22,6 +22,13 @@ setup_variables() {
         shift
     done
 
+    set +u
+    if [[ -z "$ARCH" ]]; then
```
nathanchance, line 26: Could do `if [[ -z "${ARCH:-}" ]]; then` instead of the `set +u` dance.

driver.sh
```diff
@@ -332,26 +339,38 @@ check_objcopy_strip_version() {
...
-        LD="${LD}" \
-        LLVM_IAS="${LLVM_IAS}" \
-        NM="${NM}" \
-        OBJCOPY="${OBJCOPY}" \
-        OBJDUMP="${OBJDUMP}" \
-        OBJSIZE="${OBJSIZE}" \
-        READELF="${READELF}" \
-        STRIP="${STRIP}" \
-        "${@}"
+    if [[ ${LLVM} -eq "1" ]]; then
```
nathanchance, line 342: Will probably need something like `${LLVM:-0}` due to `set -u`.

.travis.yml
```diff
@@ -24,8 +24,8 @@ jobs:
       env: ARCH=riscv LLVM_IAS=1
     - name: "ARCH=s390 BOOT=false"
       env: ARCH=s390 BOOT=false
-    - name: "ARCH=x86 LD=ld.lld"
-      env: ARCH=x86 LD=ld.lld
+    - name: "ARCH=x86 LLVM=1"
```
nathanchance, line 27: Let's get #302 merged then switch over as much as possible to `LLVM=1` + GNU variants as we need them.

driver.sh
nickdesaulniers: `./driver.sh: line 25: ARCH: unbound variable`
am I holding it wrong? 🤡

nickdesaulniers: maybe default to zero and compare against a value of zero, rather than use `-z`?

driver.sh
nickdesaulniers: yep, done

.travis.yml
nickdesaulniers: will do

driver.sh
nathanchance: Yes, that is probably a better idea.

nickdesaulniers: done

PR #302, opened 2020-08-20
add patch for stpcpy

nickdesaulniers:


nickdesaulniers:
not the right patch...one minute

PR #305, opened 2020-08-26
Revert "travis: Disable booting for RISC-V on linux-next for now"

nathanchance:
This reverts commit 590e48344c13e5910021b7cd1268dd93b8ab96b7.

This has been fixed: https://github.com/ClangBuiltLinux/linux/issues/1092

Unfortunately, the build will still be broken; still pending
investigation. I am only doing this so that I do not forget.


msfjarvis (approved):

nickdesaulniers:
so WIP then?

nathanchance:
This can go in now. I will try to look into the build failure today or tomorrow. For what it's worth, my local builds are fine so this might be something up with the specific versions of the tools that we are using.

PR #307, opened 2020-08-28
RISC-V fixes

nathanchance:
This pull "fixes" RISC-V builds and should make the build fully green.

Presubmit: https://travis-ci.com/github/nathanchance/continuous-integration/builds/181740252

The first two patches workaround https://github.com/ClangBuiltLinux/linux/issues/1143 and https://github.com/ClangBuiltLinux/linux/issues/1144. `BOOT=false` is necessary because the QEMU shipped in Ubuntu/Debian appears to be bugged as stock QEMU 4.2.0 and 5.0.0 work fine. We could probably start building QEMU in our Docker image but that is a significant amount of overhead and yet another binary that we have to triage issues with.

An alternative to this is just dropping RISC-V support for the time being until the architecture has matured a bit more. I do not personally care either way but we should really get the whole matrix fully green at this point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ jobs:
       env: ARCH=ppc64
     - name: "ARCH=ppc64le LD=ld.lld"
       env: ARCH=ppc64le LD=ld.lld
-    - name: "ARCH=riscv LLVM_IAS=1"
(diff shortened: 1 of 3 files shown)


msfjarvis (commented):
Changes look good to me, the QEMU thing is rather unfortunate though.

PR #310, opened 2020-09-02
Fix android-mainline patch application and enable LLVM=1 LLVM_IAS=1

nathanchance:
Patches should be fairly self explanatory.

Presubmit: https://travis-ci.com/github/nathanchance/continuous-integration/builds/182477086


msfjarvis (approved):

nickdesaulniers (approved):
thanks, we recently enabled 32b x86 builds for Android too recently

PR #313, opened 2020-09-13
patches: llvm-12: Refresh x86_64 linux-next patch

nathanchance:
We get conflicts after
https://git.kernel.org/tip/597cfe48212a3f110ab0f918bf59791f453e65b7.

Presubmit: https://travis-ci.com/github/nathanchance/continuous-integration/builds/184073027


msfjarvis (approved):

nickdesaulniers:
can you test dropping that patch? I don't think we should need it on next.  @nivedita76 landed a 6 patch series IIRC that addresses the assembler relaxation.  I still need to send that one liner patch to stable (based on this past week's meeting as suggested by @broonie and @arnd).

nathanchance:
Sure. Are we just planning to have it go to stable only rather than mainline first? We have tried pinging the x86 maintainers several times at this point so I am not sure what else to do.

nivedita76:
Pinged them again. It's also needed for v5.9.

There are also some trivial conflicts with the -stable trees, so we might need to send out separate patches that apply cleanly.

nickdesaulniers:
https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?h=x86/urgent&id=09e43968db40c33a73e9ddbfd937f46d5c334924

nickdesaulniers (approved):
thanks! 🍹 🔐

PR #316, opened 2020-09-21
remove x86 relaxation patch

nickdesaulniers:
fixes
https://travis-ci.com/github/ClangBuiltLinux/continuous-integration/jobs/388047047

Landed in
09e43968db40c33a73e9ddbfd937f46d5c334924.

https://github.com/ClangBuiltLinux/linux/issues/1121


nickdesaulniers:
wait, I just got emailed that this got picked up for 5.8, 5.4, and 4.19. Let me drop it for those branches, too.

nathanchance:
> wait, I just got emailed that this got picked up for 5.8, 5.4, and 4.19. Let me drop it for those branches, too.

I do not think that those kernels will be released and available by the time that our CI runs tonight. We should keep the second patch separate for a follow up PR if we want green tonight; otherwise, we can just merge this and it will go green when those come out.

nickdesaulniers:
dropped stable patches; ready for rereview

msfjarvis (approved):

